PR #6, opened 2018-09-20
fix(issuer): clearly display rate for STO

sajclarke:
Implement numeric validator and thousandsDelimiter to display rate


josepmc:
This one should be merged

PR #8, opened 2018-09-20
fix(issuer): fix trim on confirmemail page

sajclarke:
Remove trim nomralize and add type of email to reduxform field to avoid cursor position change


josepmc:
This one should be merged

PR #12, opened 2018-09-21
fix(issuer): display dummy page for incompatible browsers

sajclarke:
display dummy page for incompatible browsers


packages/polymath-issuer/src/RouteLoader.js
```diff
@@ -24,7 +24,7 @@ type Props = {
 
 class RouteLoader extends Component<Props> {
   render() {
-    if (isMobile) {
+    if (isMobile || (!isChrome && !isFirefox && !isOpera)) {
```
RafaelVidaurre: Just to be more clear, it would be nice to name the result of the second part of this statement into its own variable as well. 

So, for example, the following alternative might help so that the next person who comes understands right away what the code does:

```
render() {
  const isUnsupportedBrowser = !isChrome && !isFirefox && !isOpera
  if (isMobile || isUnsupportedBrowser) {
    // ...
  }
}
```

sajclarke: Agreed

josepmc:
This one should be merged

RafaelVidaurre:
@josepmc maybe this is something we want to check before merging then

RafaelVidaurre:
@josepmc I guess what you are saying would make sense for `browser <= ie8`, should we care about this?

josepmc:
I think it'd be wise for us to avoid any possible mishaps by an external library and just insert any snippet before anything, that would cover us in any situation.

Grsmto:
@RafaelVidaurre I think you can merge this. We have an issue already in place for what Josep suggested.

RafaelVidaurre:
@Grsmto thanks, I'll do that then

PR #14, opened 2018-09-21
Add hot reloading support + pkg.json deps cleanup

Grsmto:
- Add support for hot reloading
How to test: change a line of text or css file, content should change **without** reloading the entire page
- Reorganise some packages to dev dependencies
- Remove unused packages: `react-dropzone`, `resolve` (already a dependency of Babel), couple of Babel plugins that are already included in `babel-preset-react-app`


Grsmto:
So the CRA webpack config defines 2 separate Babel config for package's own code and external node_modules packages.
Applying RHL globally generates the error referenced above, so we apply it only for the package's own code.
Until we find a solution, RHL should work for the code within the package.

RafaelVidaurre:
Ok, will review in a bit. So, expected behavior now is:
- Hot reload works within `issuer` only
- Saves on dependencies cause browser reload

@Grsmto is this correct?

Grsmto:
Yes that's right.

PR #40, opened 2018-09-26
Fix/auth move node urls to env vars

RafaelVidaurre:
* Moves `polymath-auth`'s hard-coded URLs into env variables.
* Updates documentation

**IMPORTANT:** If you are planning on using `Kovan` locally, you **must** set the `REACT_APP_NETWORK_KOVAN_WS` env variable in the proper app pointing to the node's URL (I'll share this to you privately).

Also, note that this will most likely change soon, this is just so that we can keep working with the updated node urls

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,11 @@
 
 This is a monorepo containing the webapps and libraries in Polymath. In each package you will find a _Readme_ that describes it.
 
+# Apps and packages
+
+- [Polymath Issuer](https://github.com/PolymathNetwork/polymath-apps/blob/develop/packages/polymath-issuer/README.md)
+- [Polymath Investor](https://github.com/PolymathNetwork/polymath-apps/blob/develop/packages/polymath-investor/README.md)
+
 # Setup
 
 In the root of the repository run:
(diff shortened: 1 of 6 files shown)


packages/polymath-auth/src/networks.js
```diff
@@ -27,11 +26,10 @@ export default (id: string = 'local'): Network =>
...
     [NETWORK_KOVAN]: {
       name: 'Kovan Testnet',
-      url:
-        'wss://heartily-internal-escargot.quiknode.io/46df7525-5518-428e-b30b-9dea09480213/4oM662IRx_2tZJq3ht4wdQ==/',
+      url: process.env.REACT_APP_NETWORK_KOVAN_WS,
     },
     local: {
       name: 'Localhost',
-      url: 'ws://localhost:8545',
+      url: process.env.REACT_APP_NETWORK_LOCAL_WS,
```
josepmc: For this one, I'd suggest to do something like `process.env.REACT_APP_NETWORK_LOCAL_WS || 'ws://localhost:8545'`

PR #41, opened 2018-09-27
Feature/reorganise folder structure (fix #16)

### LINKED ISSUE #16: App folder structure organisation
Hi guys,

Creating this issue to document/discuss the app structure for `polymath-issuer` and `polymath-ui` packages.

### Issuer
Regarding the issuer app, here is the structure I'm aiming for:

screen shot 2018-09-25 at 14 59 00

- `components/`
    - Here we have `layout/` components: like `<App>`, `<Root>`, ...
    - `utils/` any other components useful for the app itself but that can't be abstracted to the `polymath-ui` component library. For example in there we have `<NotFoundPage>`.
- `pages/`
We follow the `pages` pattern, that's usually there because of code-splitting and each `page` is an app entry chunk (in Next.js apps for example).
In there you have each page usually following the router structure. These are usually **container** components (linked to Redux store).
- Any setup/routing files can stay at the root.
- `style.scss` this file should have **only** global styles. Anything targeting a specific page/component is wrong and should be refactored (ie: everything in this file...).

### Component library (`polymath-ui`)
We move away from:
```
src/
  components/
  modules/
```

to 

```
src/
  components/
```

`modules/` is where the Redux connected components and their state are located.
"modules" doesn't mean anything here so it was hard to know where things should go. Instead we get every components in a single folder.
It can make sense for some components to export a Redux state but I think in most scenarios it should be avoided as much as possible as it binds them to Redux and forces user to use Redux.
The important thing is to have dumb reusable components. Ideally we should keep any "redux" components separated from the dumb component.

The goal is to get a `polymath-ui` component library that is:
- Reusable: I can take a component and get it working on different apps
- Composable: bigger components are composed of smaller other components

Observations: 
- Most of the code from `polymath-ui` is just there to share code between `polymath-issuer` and `polymath-investors`.
The problem is that some implementation details are "shared" whereas they shouldn't. It's pushing the DRY principle too far imo. `PolymathUI.js` component is a perfect example of that.

- Redux is mostly used as a global variable to communicate between packages. A lot of the Redux state could be replaced by props.

Grsmto:
Hey guys,
Please find details about each change + the reasoning behind each of them.

### Polymath-ui

- Reorganised components (`polymath-ui/src/components/`) in folders so we can have extra files (doc, tests, children, etc.):
```
Component/
  index.js
  Component.js
  Component.md
  Component.test.js
```

- Merged `components/` and `modules/` together. This is the first step of a 2 steps task. Next step is to separate the dependency to Redux and move code that isn't related to the component itself in the `redux/` folder.
To be clear on this: the reason why there was this `modules/` folder is because these are the components that share a Redux state (used by issuer and investors apps). We have to keep these Redux actions/reducer within the `polymath-ui` repo because this is used by both dapps.
**So the `polymath-ui` repo will not/can't contain only pure components but also some Redux states used in both issuer and investor dapps.**
- Merged `svg/` in `images/` because svgs are images. 
Later we can refactor these as well because at the moment they are hardcoded as React components SVG but we could use a loader that does that automatically instead (CRA v2 has this built-in for ex).
- Removed [`PolymathUI.js` component.](https://github.com/PolymathNetwork/polymath-apps/blob/develop/packages/polymath-ui/src/PolymathUI.js) This was shared between investors and issuer apps. This simply work because investors/issuer apps are 1:1 and it's a really fortunate scenario. That wasn't flexible.
Now both apps have their own `PolymathUI.js`. Later I plan to remove that file entirely and get it merged with `App/Root` components as this is what it is.

### Polymath-auth

- Entire package removed and merged/renamed.
This repo was just a UI component. So it's now in `polymath-ui` with other components.
I renamed it `EthNetworkWrapper` as all it does is conditionally rendering components based on the Ethereum network provided as prop. 
You're welcome to propose a new name haha.

### Polymath-issuer

- Instead of a "modules" based organisation, we moved to a split by file type. The problem with the previous organisation is that it was split by "pages/section" but the Redux state is usually not coupled to a single page but is supposed to be global across the app (it's often related to a page but not necessarily).

### Polymath-investors

- Nothing changed. Just refactoring of imports due to dependencies changes.


packages/polymath-issuer/src/actions/token.js
```diff
@@ -14,12 +14,14 @@ import type {
   Address,
 } from '@polymathnetwork/js/types';
 
-import { formName as completeFormName } from './components/CompleteTokenForm';
-import { fetch as fetchSTO } from '../sto/actions';
-import { PERMANENT_LOCKUP_TS } from '../compliance/actions';
-import CreatedEmail from './components/CreatedEmail';
-import type { GetState } from '../../redux/reducer';
-import type { ExtractReturn } from '../../redux/helpers';
+// TODO: Here we are creating components inside a Redux action file...
```
RafaelVidaurre: Two things about this:
- How about we agree on always signing TODO/FIXME comments: `// TODO @RafaelVidaurre: foo` just so that we can always know who to ask for more details
- When doing this it would be even better if we also add what needs to be done (given that it is a `TODO` comment). The other option I guess is to use a `FIXME` which makes it more clear that what we are doing is wrong

packages/polymath-ui/src/components/Navbar/Navbar.js
```diff
@@ -21,6 +24,8 @@ type StateProps = {|
 |};
 
 const mapStateToProps = (state: RootState): StateProps => ({
+  // TODO: refactor this! We have a cross dependency here, this depends
```
RafaelVidaurre: This `TODO` fits perfectly with what I mentioned before in this review. Let's just add the author's github handle to it

RafaelVidaurre:
NOTE: This breaks https://github.com/PolymathNetwork/polymath-apps/pull/40 but that was expected. I'll just rebase the PR after this is merged

RafaelVidaurre:
By the way, about the `EthNetworkWrapper`, I agree we can find a better name. That being said, leave it like this for now. I think that whole component needs to be reevaluated anyways, so we can can come back to it in the near future

Grsmto:
@RafaelVidaurre thanks for the review.
I rolled back the babel-core version. Shouldn't be done in this PR anyway plus it wasn't in sync with the other packages.

PR #44, opened 2018-09-27
fix(issuer): export csv without popup

sajclarke:
This removes the need to open a popup window to download the csv export.


RafaelVidaurre:
Builds are failing here because this needs a rebase btw!

PR #50, opened 2018-09-30
Feature/smart contracts documentation utility

RafaelVidaurre:
I hacked together this little tool to help document the Smart Contract's methods and events that are relevant to the dApps. Since doing this manually was a crazy amount of work and these are gonna change soon.

It reads `config/contractDocs.config.js` and generates a `CONTRACTS.md` file based on the config object it exports. It will show any contract's methods that exist in the config file (wether they have a description string or just any truthy value). It will also add all other methods to the document but will collapse them into the `other methods` section.

P.S. I didn't give the generator's code too much love since I don't think it was worth spending too much time on a utility tool that we're probably not gonna modify/use often. So I wouldn't worry too much about the code itself

### DIFF
--- a/packages/polymath-shared/babel.config.js
+++ b/packages/polymath-shared/babel.config.js
@@ -1,7 +1,0 @@
-const base = require('../../config/babel.config.js');
-
-module.exports = {
-  babelrcRoots: ['../packages/*'],
-  presets: ['react-app'],
-  plugins: [...base.plugins],
-};
--- a/packages/polymath-shared/babel.config.js
+++ b/packages/polymath-shared/babel.config.js
@@ -1,0 +1,7 @@
+const base = require('../../config/babel.config.js');
+
+module.exports = {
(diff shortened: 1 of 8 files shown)


packages/polymath-shared/babel.config.js
```diff
@@ -0,0 +1,7 @@
+const base = require('../../config/babel.config.js');
+
+module.exports = {
+  babelrcRoots: ['../packages/*'],
```
Grsmto: I think this line is useless/wrongly written as it should be `['../*']` or I'm missing something?

Grsmto:
I tested the script and it worked just fine. 👍 
I was just wondering why you added the babel.config.js as nothing in this package is being transpiled? Is it for future-proof?

packages/polymath-shared/babel.config.js
RafaelVidaurre: Good catch, this us outdated code 💀

PR #64, opened 2018-10-04
fix(issuer): fix image links and add new provider

sajclarke:
Fix image links to provider logo and background(in modal view)
Add new provider (Messner Reeves LLC)


RafaelVidaurre:
Woah, why were those routes like this? 
Could it have been that in some mass-replace I did while migrating to the monorepo I broke those? 😨

RafaelVidaurre:
@sajclarke approved, can you solve conflicts?

PR #66, opened 2018-10-04
Feature/add snapshot testing

Grsmto:
Add support for snapshot testing on Polymath-UI.
Just a few components have been tested as poc. We will add missing tests incrementally.

The Azure build should test everything that this PR involves (so if it's green, it should means nothing else to be reviewed, as build passes and tests as well).


RafaelVidaurre:
@Grsmto is this working on your machine?

Not sure why but the `yarn test` command on root at least gets stuck forever

RafaelVidaurre:
According to what has been discussed, is seems that this was related to not having `CI=true` in Azure Devops.

RafaelVidaurre:
Closing in favor of #77 

Decided not to write over this PR for reference

PR #77, opened 2018-10-10
Fix/unit tests

RafaelVidaurre:

This PR is a continuation from #66

## Unit tests
- Fixes snapshot tests rendering null results
- Replaces Date mocking with vanilla mocking through jest's utilities
- Moves tests to `__tests__` for consistency with existing tests and cleaner directories
- Fixes issues hanging tests on `ui`
- Removes usage of `Provider` wrappers supplying a mock store. Components shouldn't require being connected to the store, we should be mocking any interactions for better isolation and testability (redux's docs also recommend this approach)
- Updates jest version:
  * Improves snapshots
  * Adds `--detectOpenHandles` to find processes hanging the tests

## Coverage
Generates all requested coverage / test report files


packages/polymath-ui/src/components/EthNetworkWrapper/__tests__/EthNetworkWrapper.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import { EthNetworkWrapper } from '../EthNetworkWrapper';
```
Grsmto: Export components and test them "not Redux connected" is a solution but I wouldn't say it's "the best solution" to the point of having to refactor this (see [comment of Enzyme author saying you shouldn't do it](https://github.com/airbnb/enzyme/issues/1002#issuecomment-340065900)).
I usually mock the Redux store <Provider> because I can also provide initial state in my tests to get snapshots of different scenarios (component with different initial props) and all of this going through Redux mapStateToProps just like on the real app.
It also allows you to snapshot an entire page without having bugs if a descendant component is connected (for certain components that you don't want to shallow render).
Overall I think it's fine for polymath-ui to do this but for issuer I would definitely mock the Redux store.

packages/polymath-ui/babel.config.js
```diff
@@ -1,12 +1,7 @@
...
-    ...base.plugins,
-    '@babel/plugin-proposal-object-rest-spread',
-    '@babel/plugin-transform-runtime',
-    '@babel/plugin-syntax-async-generators',
-    '@babel/plugin-transform-regenerator',
-  ],
+  babelrcRoots: ['../packages/*'],
+  presets: ['@babel/env', '@babel/flow', '@babel/react'],
+  plugins: [...base.plugins, '@babel/plugin-transform-runtime'],
 };
```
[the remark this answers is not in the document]
RafaelVidaurre: This is just part of all the changes I had to do to get this to build. I started adding what we needed for polymath ui from scratch so that we didn't include unnecessary stuff

packages/polymath-ui/src/components/Countdown/__tests__/Countdown.test.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import Countdown from '../Countdown';
+
+describe('Countdown', () => {
+  test('renders without crashing', () => {
+    jest
+      .spyOn(Date, 'now')
+      .mockImplementation(() => new Date('2018-01-01T05:12:34').getTime());
```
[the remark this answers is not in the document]
RafaelVidaurre: Yeah, I changed that in the code too. `Date()` is not the right way to get the current timestamp

packages/polymath-ui/src/components/EthNetworkWrapper/__tests__/EthNetworkWrapper.js
RafaelVidaurre: If you need to test multiple components rendered in the same test then that's fine, mocking the store sounds good.

For `Container` tests I'm not convinced by the argument (it has its good amount of downvotes even). The argument "you'd never use it on its own on production" doesn't make much sense if we talk about **unit**  tests. In unit tests we want isolation.

That being said, I think testing the component wrapped on a `Provider` component is fine if the store is mocked. So if it makes a test simpler I have no issue with that, its a matter of trade-offs. One requires mocking the HOCs the other requires exporting the different pieces of *our* code only and mock the APIs.

Bottomline:
**Let's do whatever is simple as long as we are able to test properly and tests are not too fragile**

P.S: I removed the global mocked store until we actually need something like that, I do think that could be dangerous if it is not reset after each test, and a utility function to me feels safer than a global object shared across tests (for instance, what happens if tests run in parallel and mutate this store?).

packages/polymath-ui/package.json
```diff
@@ -36,36 +37,65 @@
...
     "redux-thunk": "^2.3.0",
     "regenerator-runtime": "^0.12.1",
     "sass-loader": "^6.0.7",
     "style-loader": "^0.20.3",
-    "webpack": "^4.19.0",
-    "webpack-cli": "^3.1.0"
+    "webpack": "^4.19.0"
   },
   "jest": {
+    "testMatch": [
```
josepmc: Jest config is different in ui than in offchain and issuer. Is this intended?

PR #81, opened 2018-10-10
fix(issuer): update datetime pickers on STOConfigureForm

sajclarke:
Update datetime picker components on STOConfigureForm


packages/polymath-issuer/src/pages/sto/components/ConfigureSTOForm.js
```diff
@@ -65,9 +71,36 @@ class ConfigureSTOForm extends Component<Props, State> {
     );
   };
 
+  checkStartAfterEnd = (value, allValues) => {
+    // console.log(value, allValues.endDate[0])
```
RafaelVidaurre: Let's remove this comments

packages/polymath-issuer/src/pages/sto/components/ConfigureSTOForm.js
```diff
@@ -65,9 +71,36 @@ class ConfigureSTOForm extends Component<Props, State> {
...
+        return 'Time is in the past';
+      }
+    }
+  };
+
   pastCheck = (value, allValues) => {
     const startTime = allValues.startTime;
-    const dateRange = allValues['start-end'];
+    const dateRange = allValues['startDate'];
+    console.log(value, allValues);
```
RafaelVidaurre: Remove this too

packages/polymath-issuer/src/pages/sto/components/ConfigureSTOForm.js
```diff
@@ -82,6 +115,7 @@ class ConfigureSTOForm extends Component<Props, State> {
         hours,
         parseInt(minutes, 10)
       );
+      console.log(startDateTime);
```
RafaelVidaurre: And this ;p

packages/polymath-issuer/src/pages/sto/components/ConfigureSTOForm.js
```diff
@@ -100,14 +134,42 @@ class ConfigureSTOForm extends Component<Props, State> {
...
+            name="endTime"
+            step={30}
+            component={TimePickerSelect}
+            className="bx--time-picker__select"
+            placeholder="hh:mm"
+            label="Time"
+          />
+        </div>
+
+        {/* <div className="time-pickers-container">
```
RafaelVidaurre: And this ;p

packages/polymath-ui/src/components/inputs/DatePickerInput.js
```diff
@@ -19,34 +20,37 @@ type Props = {
...
-      id={input.name}
+      id={name}
       className={className}
       datePickerType="single"
       // eslint-disable-next-line
       onChange={(date: Date) => {
-        input.onChange(date || null);
+        onChange(date || null);
+        // redux-form updates `touched` on blur.
+        onBlur();
```
RafaelVidaurre: I'm not convinced this is the best way to do this, it is missleading to call `onBlur()` if we are not losing focus to the component.

I'll play with this component this week to have some more context and offer a better solution if there is one.

packages/polymath-ui/src/validate.js
```diff
@@ -110,6 +110,7 @@ export const todayOrLater = (date: ?Date): ?string => {
   const dayBegin = new Date(now.getFullYear(), now.getMonth(), now.getDate());
 
   if (dayBegin.getTime() > date.getTime()) {
+    console.log('Date is in the past.');
```
RafaelVidaurre: Remove this too

Grsmto:
As a reference, the styling issue on the datepicker has been fixed [there](https://github.com/PolymathNetwork/polymath-apps/pull/86/commits/e1644eab0c6c82ce11834e7dbb7e84a3a90d3379).

packages/polymath-ui/src/components/inputs/DatePickerInput.js
sajclarke: I have tried a few other options but seems this is a bug in redux-form. If you find a better way kindly let me know

sajclarke:
This was a WIP update @RafaelVidaurre so the code clean-up was not completed.
@Grsmto thanks for fixing that issue!

packages/polymath-ui/src/components/inputs/DatePickerInput.js
RafaelVidaurre: Will do @sajclarke I'll provide a solution if there is one as soon as I can.

Grsmto: I had a look at this and I think it's the only way. I think it's fine to do this because it doesn't actually blur the field. It just calls "onBlur" to Redux-form as that's the only way to trigger the validation.

PR #90, opened 2018-10-15
Fix/history missing redux state

Grsmto:

- [ ] I linked the issues related to this PR in its description (if any).

This will need refactoring later as the Polymath-ui components should not call any history change directly. At the moment the `TxModal` is triggering some routing, that should be handled by the caller's code not by this component directly (this should be done with a callback).

ref:
image


RafaelVidaurre:
Merging, thanks for fixing this

PR #100, opened 2018-10-16
Feature/smart contract addresses in code

RafaelVidaurre:

This PR moves smart contract addresses to code:
- Hardcodes addresses per `deployment stage` + `network id`
- Most of this code is only meant to exist until we can transition to 2.0
- Modifies polymath.js' smart contracts to use hardcoded addresses unless explicitly passed an address
- Sets `REACT_APP_DEPLOYMENT_STAGE` to default to `local` 
- Adds tests for `constants` expected values per environment
- Sets up jest for `shared` repo


packages/polymath-js/src/contracts/TickerRegistry.js
```diff
@@ -73,6 +73,7 @@ class TickerRegistry extends Contract {
 
   /**
    * TODO @bshevchenko: DEPRECATED since owner can be changed
+   * NOTE @RafaelVidaurre: Not really depracated, still being used
```
Grsmto:
```suggestion
   * NOTE @RafaelVidaurre: Not really deprecated, still being used
```

packages/polymath-shared/babel.config.js
```diff
@@ -0,0 +1,12 @@
+const base = require('../../config/babel.config.js');
+module.exports = {
+  babelrcRoots: ['../packages/*'],
+  presets: ['@babel/flow', '@babel/env', '@babel/react'],
```
Grsmto: `@babel/react` this stuff might fail as it's not in the package.json. But except if you plan to have some React components in this package, you could remove this.

packages/polymath-js/src/contracts/TickerRegistry.js
RafaelVidaurre: Love the new feature <3

packages/polymath-shared/babel.config.js
RafaelVidaurre: True that, we don't need this for now 👍

PR #102, opened 2018-10-16
[WIP]Metamask breaking change update

Grsmto:
- `MetamaskStatus` is now a dumb component (not connected to Redux anymore)
- Add support for Metamask authorization to `EthNetworkWrapper` component
- Add support for new Metamask version global object `window.ethereum` to `EthNetworkWrapper` component

Out of context changes:
- Add new Button component as part of `polymath-ui` (with the goal of deprecating direct usage of `carbon-react-components` from issuer and have a single component library)
- Apply same version of `carbon-components` to `polymath-ui` as `polymath-issuer`


Grsmto:
Ready for testing.

packages/polymath-issuer/src/components/Root.js
```diff
@@ -62,9 +62,15 @@ class Root extends Component<Props> {
             <SplashPage />
           ) : (
             <EthNetworkWrapper
-              loading={<Loading />}
-              guide={<MetamaskStatus networks="Mainnet or Kovan" />}
               networks={networks}
+              Loading={<Loading />}
+              errorRender={({ networkError, onRequestAuth }) => (
```
RafaelVidaurre: Renderprops = life

packages/polymath-ui/src/components/EthNetworkWrapper/EthNetworkWrapper.js
```diff
@@ -56,14 +54,34 @@ export class EthNetworkWrapper extends Component<Props> {
...
+
+    if (isFailed) {
+      return errorRender({
+        networkError: error,
+        onRequestAuth: this.handleAuthRequest.bind(this),
+      });
     }
-    if (!this.props.isFailed) {
-      return this.props.loading;
+    if (!isFailed && !isConnected) {
```
RafaelVidaurre: We can remove the `!isFailed` here
```suggestion
    if (!isConnected) {
```

packages/polymath-ui/src/components/EthNetworkWrapper/actions.js
```diff
@@ -17,85 +32,119 @@ export type NetworkParams = {|
...
-const fail = (error: number) => ({ type: FAIL, error });
+const fail = (error: number) => ({ type: FAILED, error });
 
-export type Action =
-  | ExtractReturn<typeof connected>
-  | ExtractReturn<typeof fail>;
+export const init = (networks: Array<string>) => async (dispatch: Function) => {
+  let web3;
+  let web3WS; // since MetaMask doesn't support WebSockets we need this extra client for events subscribing
+  let networkId = undefined;
```
RafaelVidaurre: This is undefined by default

packages/polymath-ui/src/components/EthNetworkWrapper/actions.js
```diff
@@ -17,85 +32,119 @@ export type NetworkParams = {|
...
 
-const web3 = new Web3();
-const web3WS = new Web3(); // since MetaMask doesn't support WebSockets we need this extra client for events subscribing
+  // Instantiate Web3 HTTP
+  if (window.ethereum) {
```
RafaelVidaurre: This seems like a good opportunity for self-documenting code.

```suggestion
export const init = (networks: Array<string>) => async (dispatch: Function) => {
  let web3;
  let web3WS;
  let networkId;
  const newProviderInjected = !!window.ethereum;
  const oldProviderInjected = !!window.web3;
  
  if (newProviderInjected) {
    web3 = new Web3(window.ethereum);
  } else if (oldProviderInjected) {
    web3 = new Web3(window.web3.currentProvider);
  } else {
     return dispatch(fail(ERROR_NOT_INSTALLED)); 
  }
}
```

packages/polymath-ui/src/components/EthNetworkWrapper/index.js
```diff
@@ -3,21 +3,25 @@
...
-} from './actions';
+
+export const ERROR_NOT_INSTALLED = 1;
+export const ERROR_LOCKED = 2;
+export const ERROR_NETWORK = 3;
+export const ERROR_DISCONNECTED = 4;
+export const ERROR_ACCESS_REQUESTED = 5;
+export const ERROR_ACCESS_DENIED = 6;
+
+export const CONNECTED = 'polymath-auth/CONNECTED';
```
RafaelVidaurre: I'm not not super convinced this is a good idea of defining/exporting code (I know it was being done this way) coming from actions in `EthNetworkWrapper/index.js`. Shouldn't these come from the `actions.js` file?

RafaelVidaurre:
An extra note @Grsmto There's a `refactor` type in conventional commits we can use (since you used chore mostly).

That being said, I still think something more minimalistic like just having `fix`, `feat` and one generic type that doesn't bump any versions like (`improve`) is better than having all these `docs`, `refactor`, `build` etc. There are too many ambiguous cases.

Maybe we should try that?

packages/polymath-ui/src/components/EthNetworkWrapper/actions.js
```diff
@@ -17,85 +32,119 @@ export type NetworkParams = {|
 // this has to be here for now
 const HARDCODED_NETWORK_ID = 15;
 
-export const CONNECTED = 'polymath-auth/CONNECTED';
+export const requestAuthorization = () => async (dispatch: Function) => {
+  try {
+    dispatch(fail(ERROR_ACCESS_REQUESTED));
```
RafaelVidaurre: This one @Grsmto Is this ok? It seems like the request authorization dispatches an error in any case, and then if `enable` doesnt work it dispatches *another* action with `ERROR_ACCESS_DENIED`

Grsmto: I understand how this is confusing. This is kind of an heritage from the previous code. As we have the polling `setInterval` that reloads the page when Metamask has changed, we don't need to have a success action triggered here because page would be reloaded anyway.
That's why there is only the error case handled here.
I refactored this btw and I'll add a comment why what I just explained.

Grsmto:
> An extra note @Grsmto There's a `refactor` type in conventional commits we can use (since you used chore mostly).
> 
> That being said, I still think something more minimalistic like just having `fix`, `feat` and one generic type that doesn't bump any versions like (`improve`) is better than having all these `docs`, `refactor`, `build` etc. There are too many ambiguous cases.
> 
> Maybe we should try that?

Yep I could have used refactor for some of these. Thanks for the reminder. I'm still getting along with conventional commits! I think it's fine like that, just need to get used to it I guess.

packages/polymath-ui/src/components/EthNetworkWrapper/actions.js
RafaelVidaurre: Awesome, thanks!

Grsmto:
>- If we deal with some part of the app with good isolation we should just refactor that to a high-quality level if that is affordable
>- If we cannot do .1 we should leave a comment indicating that this needs further refactoring and/or is legacy code.

I think the scope of the changes are actually fairly isolated so I think the risk of breaking something is fine. I start to have good knowledge of the user flow now and the way the app kickstarts so I'm confident enough about this.

>Since the Metamask breaking changes support is urgent I know this is a tough dilemma. So this is fine for merging, but since we agreed on improving small issues the code as we touch it let's either take the time to make those small changes and run small risks or leave some sort of indication that this is not ideal code yet.

Problem here is that doing this task was hard to do with the old codebase (hard to understand what was happening and where to put the new code) so I kinda needed to refactor (at a risk yes...). Also it's the right time to refactor when you're adding new code! (incrementally).

FYI Metamask is still actively working on this so that's still not stable and we should probably not merge this yet: https://github.com/MetaMask/metamask-extension/pull/4703#issuecomment-431052164

packages/polymath-ui/src/components/EthNetworkWrapper/index.js
RafaelVidaurre: At least I think these constants are more coupled to the action creators than a specific component (even if they were made explicitly for such component). This is also  what I've at least seen as common practice

RafaelVidaurre:
> FYI Metamask is still actively working on this so that's still not stable and we should probably not merge this yet: [MetaMask/metamask-extension#4703 (comment)](https://github.com/MetaMask/metamask-extension/pull/4703#issuecomment-431052164)

Woah... That's just insane, they plan to release on November 2nd but they are still adding new code two weeks before that.

packages/polymath-ui/src/components/EthNetworkWrapper/index.js
Grsmto: Yeah in this context I agree as the component doesn't make much sense anyway with these Redux actions that are not really encapsulated/tighten to the component...

Grsmto:
- Applied code review changes
- Updated code with [latest Metamask build](https://github.com/MetaMask/metamask-extension/pull/4703#issuecomment-431052164)

PR #104, opened 2018-10-17
Fix csv download

sajclarke:


josepmc:
@monitz87 was suggesting to use an iframe instead.

packages/polymath-issuer/src/actions/compliance.js
```diff
@@ -239,7 +239,13 @@ export const exportWhitelist = () => async (
         ].join(',');
     });
 
-    window.location.assign(encodeURI(csvContent));
+    //Required to trigger file download in Chrome and Firefox
+    let link = document.createElement('a');
+    link.setAttribute('href', encodeURI(csvContent));
+    link.setAttribute('download', 'whitelist.csv');
+    document.body.appendChild(link);
+    link.click();
```
Grsmto: This would be a massive security issue if this was possible! Imagine triggering a download from the code :p

sajclarke:
Updated PR to use the Filesaver library

RafaelVidaurre:
@josepmc everything good from your side for the updated version of this PR?

PR #105, opened 2018-10-17
Support core 2.0 breaking changes

monitz87:

This PR handles all core 2.0 breaking changes in our current apps


packages/polymath-js/src/contracts/Contract.js
```diff
@@ -15,8 +15,17 @@ import type {
 // this has to be here for now
 const HARDCODED_NETWORK_ID = 15;
 
+const noRegistryAddressError = new Error(
```
RafaelVidaurre: Just in case, not sure what happens here with the stack trace. If the stack trace is set when the error is instanced then it'd be better to just save the text in the constant and instance the new error where we throw it.

packages/polymath-js/src/contracts/Contract.js
```diff
@@ -70,30 +111,21 @@ export default class Contract {
     ).eth.Contract(this._artifact.abi, this.address);
   }
 
-  /** @private */
-  _init(at?: Address) {
+  /**
+    Sets address and instantiates web3 contract
```
RafaelVidaurre: Not too relevant since this whole library will be replaced, but just for the future. Usually comments like these are done in this format:

```suggestion
    /**
     * Sets address and instantiates web3 contract
     * ... stuff
     */
```

packages/polymath-offchain/src/emails/STOScheduled.js
```diff
@@ -92,7 +92,7 @@ export const STOScheduled = ({
           </p>
         </div>
         <p align="center">
-          <a href={`${POLYMATH_ISSUER_URL}/dasbhoard/${ticker}/sto`}>
```
RafaelVidaurre: LOL

packages/polymath-js/src/contracts/Contract.js
monitz87: My editor doesn't do it 😢 

I did it by hand though

PR #112, opened 2018-10-18
Feature/remove url trailing slashes

monitz87:

This PR fixes a bug caused by trailing slashes in URL env variables. The composed urls for images and links in emails sent by offchain were faulty and caused images not to render and links to not redirect correctly. This, in turn, caused the emails to be marked as spam.

I implemented a small utility function (and with it, an utils model) in polymath-shared for this. We should use it for all URL composition. There are a couple of caveats regarding relative vs absolute paths which are addressed in the function docs.

### DIFF
--- a/packages/polymath-offchain/src/emails/EmailWrapper.js
+++ b/packages/polymath-offchain/src/emails/EmailWrapper.js
@@ -3,6 +3,7 @@
 import React from 'react';
 import { POLYMATH_OFFCHAIN_URL, NETWORKS } from '../constants';
 import styles from './styles';
+import { composeURL } from '@polymathnetwork/shared/utils';
 
 import type { Node } from 'react';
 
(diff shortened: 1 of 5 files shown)


RafaelVidaurre:
Hmm, to be honest I would've just made a utility to clean trailing slashes. The only place where we have uncertainty of them is from env variables. This update means we have to remember to use this in every place where we build a URL from one of our apps.

What do you think about doing something like this @monitz87 ?

- Implement a small helper that removes trailing slashes from URLs if they exist
- Always cleanup all URLs coming from `process.env` in `constants.js`

```js
// constants.js
const SOME_URL = normalizeURL(process.env.SOME_URL);
```

Now we can just *trust* the URLs in `constants.js` to never have a trailing slash.

monitz87:
Sure, I'll hack it up later

monitz87:
*Update:* this PR now implements the utility function as specified in @RafaelVidaurre 's comment

PR #117, opened 2018-10-22
Fix transactions failing in staging

monitz87:

This PR addresses https://app.asana.com/0/847528882697263/873577594811143

The script that deploys contracts assumes that PolyToken exists in kovan and hardcodes the address into the PolymathRegistry (this makes sense in production to maintain balances and other data, but in staging we would want a separate PolyToken instance every time we deploy). For this reason, for now we had to use that address in our staging kovan environment. It also includes a bunch of minor tweaks and fixes.

### DIFF
--- a/packages/polymath-js/src/contracts/Contract.js
+++ b/packages/polymath-js/src/contracts/Contract.js
@@ -200,7 +200,7 @@ export default class Contract {
     const params = {
       ...preParams,
       gas,
-      gasPrice: 5000000000,
+      gasPrice: 10000000000,
     };
 
     // dry run
--- a/packages/polymath-js/src/contracts/TickerRegistry.js
+++ b/packages/polymath-js/src/contracts/TickerRegistry.js
@@ -119,7 +119,7 @@ class TickerRegistry extends Contract {
         swarmHash
       ),
       null,
-      1.15
+      1.05
     );
   }
 }
--- a/packages/polymath-offchain/src/startup/index.js
+++ b/packages/polymath-offchain/src/startup/index.js
@@ -17,14 +17,18 @@ import './setupMailing';
 import connectWeb3 from './setupWeb3';
 
 import type { NetworkOptions } from '../constants';
+import type { NetworkId } from '@polymathnetwork/shared/constants';
 
 /**
   Attemps to connect to a network via web3.
 
-  @param {string} networkId id of the network
+  @param {NetworkId} networkId id of the network
   @param {NetworkOptions} options connection options
  */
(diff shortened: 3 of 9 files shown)


packages/polymath-js/src/contracts/Contract.js
```diff
@@ -200,7 +200,7 @@ export default class Contract {
     const params = {
       ...preParams,
       gas,
-      gasPrice: 5000000000,
+      gasPrice: 10000000000,
```
RafaelVidaurre: What are the implications of this change in production?

packages/polymath-js/src/contracts/TickerRegistry.js
```diff
@@ -119,7 +119,7 @@ class TickerRegistry extends Contract {
         swarmHash
       ),
       null,
-      1.15
+      1.05
```
RafaelVidaurre: Is it safe to lower this buffer?

packages/polymath-offchain/src/startup/index.js
```diff
@@ -44,6 +48,8 @@ const connectToNetwork = async (networkId: string, options: NetworkOptions) => {
     if (retries >= maxRetries) {
       if (optional) {
         logger.warn(message);
+        break;
+        q;
```
RafaelVidaurre: I'm guessing this `q;` was accidental ;p

packages/polymath-js/src/contracts/TickerRegistry.js
monitz87: That was the original value, I changed it some PRs ago to match the CLI, but all other calls to `_tx` have `1.05`

packages/polymath-js/src/contracts/Contract.js
monitz87: People will be charged double gas price for their transactions but they will mine faster. Most transactions are under 10 cents normally anyway. Maybe it's something we should consult with the product team first?

packages/polymath-offchain/src/startup/index.js
monitz87: Most definitely

packages/polymath-js/src/contracts/Contract.js
monitz87:
```suggestion
      gasPrice: 5000000000,
```

packages/polymath-js/src/contracts/TickerRegistry.js
RafaelVidaurre: Gotcha

PR #125, opened 2018-10-24
fix(ui): fix footer links

Grsmto:
remove usage of React Router as they are external links.


monitz87:
Everything looks good. I tested it locally and it works just fine. Gogo merge!

PR #126, opened 2018-10-25
fix(ui): truncate notice messages if they are too long

monitz87:

This PR fixes the notice bar layout being broken if the notice content is too long. It does this by using flexbox to make the bar a bit more responsive and using a javascript library to clamp the text to a maximum of 2 lines (adding ellipses at the end if necessary). There is no pure-CSS way of managing multi-line clamping, so the library is a necessity for now.


packages/polymath-ui/src/components/NoticeBar/NoticeBar.js
```diff
@@ -3,6 +3,8 @@
 import React, { Component } from 'react';
 import { Icon } from 'carbon-components-react';
 import { connect } from 'react-redux';
+import clamp from 'clamp-js-main';
+import ReactDOMServer from 'react-dom/server';
```
RafaelVidaurre: Why was this dependency added?

Grsmto:
Would be sweet to have an .mdx for this (but then we need to extract the Redux part of the component and make it dumb) so we can test it easily.
However if this issue is urgent we can do that later I think.

packages/polymath-ui/src/components/NoticeBar/NoticeBar.js
```diff
@@ -50,24 +52,50 @@ class NoticeBar extends Component<Props> {
...
   }
+
+  componentDidUpdate() {
+    /**
+     * NOTE @monitz87: I had to use this fork of the 'clamp-js' library (https://www.npmjs.com/package/clamp-js-main)
+     * because line clamping is horribly unsupported in CSS (and what little support there is is very browser-specific).
+     * What this code does is truncate the title and paragraph (maintaining the integrity of any HTML elements inside it) if they
+     * exceeds 2 lines, adding ellipses to replace the missing text
+     */
+    const paragraph = document.getElementsByClassName('pui-notice-bar-text')[0];
```
RafaelVidaurre: We should use a Ref here instead of a document selector.

RafaelVidaurre: Also is there a way we can jut pass this in render? Feels like we are falling out of the React flow

packages/polymath-ui/src/components/NoticeBar/NoticeBar.js
monitz87:
```suggestion
```

packages/polymath-ui/src/components/NoticeBar/NoticeBar.js
monitz87: The way clamp works means it has to be done post-render AFAIK. But yeah, refs are a good idea

PR #134, opened 2018-11-05
fix(issuer): redirect to /ticker on signin fail

Grsmto:
As a ref here is the original issue:

> Steps to Reproduce (incl. in Production)
> - Navigate to a token-specific url (e.g. localhost:3000/dashboard/TEST-XXX)
> - Change to a metamask account that doesn't hold an account with Polymath (e.g. didn't register and verify her/his email)
> - Sign and create an account
> 
> Actual Result
> After creating an account you're redirected to a 404 page.
> 
> Expected Result
> You should be shown the email verification screen.

### DIFF
--- a/packages/polymath-issuer/src/components/App.js
+++ b/packages/polymath-issuer/src/components/App.js
@@ -80,6 +80,13 @@ class App extends Component<Props> {
     this.props.signIn();
   }
 
+  onAuthFail() {
+    // Make sure user is on the ticker page if he doesn't have an account yet
+    if (this.props.location.pathname !== '/ticker') {
+      this.props.history.push('/ticker');
+    }
+  }
+
   render() {
     const { ticker, isFetching, route } = this.props;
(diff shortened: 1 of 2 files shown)


packages/polymath-issuer/src/components/App.js
```diff
@@ -91,7 +98,9 @@ class App extends Component<Props> {
         <TxModal />
         <EnterPINModal />
         <ConfirmModal />
-        <AuthWrapper>{renderRoutes(route.routes)}</AuthWrapper>
+        <AuthWrapper onFail={this.onAuthFail.bind(this)}>
```
RafaelVidaurre: Mind if we agree on not use bind and use class properties instead?

```js
class Foo {
  onAuthFail = () => {} // This will be binded

  render() {
    // code
    return <Foo onFail={this.onFail} />
  }
}
```

PR #139, opened 2018-11-06
Fix/kovan registry address simplification

monitz87:

This PR removes the need to append the stage to the kovan polymath registry address env vars in both issuer and offchain. Now we should have only `POLYMATH_REGISTRY_ADDRESS_KOVAN` in offchain and `REACT_APP_POLYMATH_REGISTRY_ADDRESS_KOVAN` in issuer, and their values will depend on the deployment stage


RafaelVidaurre:
@josepmc We need to update the env vars for this now right?

PR #151, opened 2018-11-08
Epic/2.0 contracts

monitz87:

This PR fixes a bug when launching investor caused by a wrong version of react-dev-utils and also adapts calls to ST's verifyTransfer to the latest polymath-core version. For now, the last parameter isn't used, so passing 0x00 to it is the way to go (this is what they're doing in the CLI)


monitz87:
Wrong branches, had to close

PR #158, opened 2018-11-09
fix: metamask not reloading on authorize

RafaelVidaurre:
**Please make sure the following boxes are checked before submitting your Pull Request:**

- [X] I linked the issues related to this PR in its description (if any).
- [X] If this PR adds new code that is not going to change in the near future, it includes unit tests to cover it.

# This PR
Reloads the page after authorizing metamask and fixes the "Connect to metamask" button

PR #159, opened 2018-11-09
fix(ui): make sure user is privacy mode off before asking for permission

Grsmto:
Invalidates #158

### DIFF
--- a/packages/polymath-ui/src/components/EthNetworkWrapper/actions.js
+++ b/packages/polymath-ui/src/components/EthNetworkWrapper/actions.js
@@ -118,7 +118,7 @@ export const init = (networks: Array<string>) => async (dispatch: Function) => {
   if (
     !accounts.length &&
     newProviderInjected &&
-    window.ethereum._metamask.isApproved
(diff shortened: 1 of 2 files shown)


packages/polymath-ui/src/components/EthNetworkWrapper/actions.js
```diff
@@ -115,7 +115,11 @@ export const init = (networks: Array<string>) => async (dispatch: Function) => {
   });
 
   // Check if dapp is authorized by Metamask/Mist
-  if (newProviderInjected && window.ethereum._metamask.isApproved) {
+  if (
+    !accounts.length &&
+    newProviderInjected &&
+    window.ethereum._metamask.isApproved
```
RafaelVidaurre: We might also want to check for the existence of `_metamask` since that object is non standard. This unless we are sure we are using Metamask in this part of the code

RafaelVidaurre:
```suggestion
    window.ethereum._metamask && window.ethereum._metamask.isApproved
```

PR #176, opened 2018-11-14
Feat/increase token deployment gas price

monitz87:

This PR increases the base gas price for token deployment from the one calculated by metamask to 15 gwei. It adds an extra parameter to the `_tx` method in the `Contract` class to set a gas price in gwei for this purpose


RafaelVidaurre:
Gotcha, thanks for the clarification

PR #180, opened 2018-11-14
fix(issuer): redirect to token freezing page appropriately

monitz87:

This PR redirects the issuer to the "Freeze and Renounce ownership" page only if they try to access the dashboard of a 1.3 token they own


packages/polymath-issuer/src/components/MigrateTokenPage.js
```diff
@@ -149,14 +170,14 @@ class MigrateTokenPageContainer extends Component {
...
-class MigrateTokenPage extends Component {
+class MigrateTokenPage extends Component<{| handleClick: Function |}> {
   render() {
     return (
       <DocumentTitle title="Migrate your Token – Polymath">
         <div className="pui-single-box">
           <div className="pui-single-box-header">
             <div className="pui-single-box-new-version-bull">
-              <img src="2.0-logo.svg" alt="Bull" />
+              <img src="/2.0-logo.svg" alt="Bull" />
```
RafaelVidaurre: You should be able to import the image as a module. This also prevents this from breaking if the name is changed during compilation for production.

RafaelVidaurre: I know this was not a change from this PR but this might break somewhere otherwise

PR #253, opened 2018-12-07
Fix/unit tests exit code

RafaelVidaurre:

# This PR

- Fixes issue with lerna throwing exit code 1 since some test commands didn't exist
- Removes Modal snapshot test breaking because of missing env vars (we need to look at this in more detail, but since it was only a snapshot test for legacy I removed it to get this working)
- Edit tests using `react-test-renderer` to use the `render` wrapper on `testUtils` which works with `react-testing-library`. The other form of testing was having issues with unmounted components. (Also for consistency)


RafaelVidaurre:
@Grsmto not really, it was a warning that some env vars were not set (they shouldn't be required for the test environment, so something's wrong there). It probably is working for you because you have a .env file somewhere preventing the error from happening, this should probably be fixed but since we won't be adding much more unit tests to the old code it is probably a bad use of our time right now

PR #267, opened 2018-12-18
Feature/typescript polymath ui

Grsmto:
This PR includes the `ModalTx` presenter.


RafaelVidaurre:
Improvements I'd suggest:

- Favor absolute imports for parents, no `../../../helpers` use `~/helpers` instead
- Move `FormikContext` to its own file to avoid non-fatal circular dependency
- Consider moving styles to their own separate files, you have the final say on this one, but it kinda bloats the markup IMO

Overall typing fixes you can apply in the meantime:
- Use `FC` from `react` to type function components
- SVG Icons are used as strings but typed as components, check what the loader actually does


Some stuff I did that you can do on other files that are failing:
- Exported Props interfaces from components where errors were found, these should always be exported
- Updated children props for InputError since context uses children as a function style of props
- Added default value for createContext as required by react. Useful for testing in isolation without requiring a provider
- changed some types literals to interfaces (should always be preferred)
- Added context typing to FormItemContext as it was not typed
- Change as in Box since it was optional (defaulted to div). Is it being used somewhere btw?
- Optional props changed to required since code expects values to exist
- Moved duplicated networkIds to constants/networks' enum

PR #280, opened 2019-01-07
fix(issuer): fix layout broken on overview capped sto

Grsmto:


packages/polymath-ui/src/components/STOStatus/STOStatus.js
```diff
@@ -84,71 +85,72 @@ export default class STOStatus extends Component<Props> {
...
-          <Box mb="l">
-            <ProgressBar progress={fractionComplete / 100} />
-          </Box>
-          <div className="pui-sto-status-bottom-row">
-            <div className="pui-sto-status-dates">
-              <div className="pui-key-value">
-                <div>Start Date</div>
-                {dateFormat(details.start)}
-              </div>
+        <Grid gridTemplateColumns={['', '', '', '1fr minmax(250px, 1fr)']}>
```
RafaelVidaurre: Just a question: This empty strings would go away in the future with breakpoints in a object right?

PR #281, opened 2019-01-07
Data fetching/cache management

monitz87:

This PR implements a DataFetcher component which handles all requests to the blockchain and caching of fetched data. It also makes several improvements to the SDK as well as general typing.

It also contains WIP for the Checkpoint and Dividends page Containers, as well as supporting sagas, actions and reducers.

Unit tests are not 100% complete, those are coming once the shape of some things is more clear


packages/new-polymath-issuer/src/components/Checkpoints/Container.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { Dispatch } from 'redux';
+import { Presenter } from './Presenter';
```
RafaelVidaurre: Lets not fix this now, but this is a note to self to have a tslint enforce (hopefully auto-fix) imports from:

- Libraries
- Absolute Imports
- Relative Imports

packages/new-polymath-issuer/src/components/Checkpoints/Container.tsx
```diff
@@ -0,0 +1,34 @@
...
+export interface Props {
+  dispatch: Dispatch<any>;
+  symbol: string;
+}
+
+export class ContainerBase extends Component<Props> {
+  public render() {
+    const { symbol } = this.props;
+    return (
+      <DataFetcher
```
RafaelVidaurre: <3

PR #284, opened 2019-01-09
fix(issuer): display error when an invalid whitelist csv is uploaded

sajclarke:

### DIFF
--- a/packages/polymath-issuer/src/utils/parsers/index.js
+++ b/packages/polymath-issuer/src/utils/parsers/index.js
@@ -72,6 +72,7 @@ export function parseWhitelistCsv(file: string) {
   const data = csvParse(file, {
     skip_empty_lines: true,
     trim: true,
+    relax_column_count: true,
     cast: (rawValue, context) => {
       const value = rawValue;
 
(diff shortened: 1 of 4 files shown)


packages/polymath-issuer/src/utils/parsers/index.js
```diff
@@ -72,6 +72,7 @@ export function parseWhitelistCsv(file: string) {
   const data = csvParse(file, {
     skip_empty_lines: true,
     trim: true,
+    relax_column_count: true,
```
RafaelVidaurre: Why do we need to use this?

sajclarke: This was initially used to handle the inconsistent number of fields between the header and data rows

packages/polymath-issuer/src/utils/parsers/index.js
```diff
@@ -106,6 +107,8 @@ export function parseWhitelistCsv(file: string) {
         return new BigNumber(value);
       }
 
+      console.log(context);
```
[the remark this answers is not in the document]
sajclarke: Resolved

PR #285, opened 2019-01-09
fix(issuer): handle pasted text with non-numeric characters

sajclarke:

### DIFF
--- a/packages/polymath-ui/src/components/inputs/NumberInput/index.js
+++ b/packages/polymath-ui/src/components/inputs/NumberInput/index.js
@@ -118,9 +118,8 @@ export class NumberInput extends Component<Props, State> {
     { min, max, useBigNumbers, maxDecimals }: Props
   ) {
     let value = v;
-    console.log(typeof value);
     if (typeof value === 'string') {
-      value = value.replace(/[^0-9]+/g, '');
+      value = value.replace(/[^0-9\.]+/g, '');
     } else if (typeof value === 'number') {
       value = String(value);
     }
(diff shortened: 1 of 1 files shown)


packages/polymath-ui/src/components/inputs/NumberInput/index.js
```diff
@@ -118,8 +118,9 @@ export class NumberInput extends Component<Props, State> {
     { min, max, useBigNumbers, maxDecimals }: Props
   ) {
     let value = v;
+    console.log(typeof value);
```
RafaelVidaurre: Left a `console.log` here.

packages/polymath-ui/src/components/inputs/NumberInput/index.js
```diff
@@ -118,8 +118,9 @@ export class NumberInput extends Component<Props, State> {
...
     if (typeof value === 'string') {
-      value = value.replace(/,/g, '');
+      value = value.replace(/[^0-9]+/g, '');
```
RafaelVidaurre: What happens with `.` characters for decimals? Wouldn't they get replaced too? Btw this component is about 80% covered by tests I believe, so it might help to `yarn test --watch` them in case you break something without knowing

sajclarke: Thanks for pointing that out. I forgot that decimals are allowed now on those fields. I've resolved and tested  with the latest commit

packages/polymath-ui/src/components/inputs/NumberInput/index.js
sajclarke: Thanks. Done.

PR #297, opened 2019-01-16
[WIP] fix(issuer): make header row mandatory on whitelist

sajclarke:

### DIFF
--- a/packages/polymath-issuer/src/utils/parsers/index.js
+++ b/packages/polymath-issuer/src/utils/parsers/index.js
@@ -83,68 +83,89 @@ const checkColumnCount = data => {
   return errorMsg;
 };
 
+const checkCSVHeaders = data => {
+  // Split the input into lines
+  let rows = data.split('\n');
+  const header_row = rows[0].split(',');
+  const default_headers =
+    'Address,Sale Lockup,Purchase Lockup,KYC/AML Expiry,Can Buy From STO,Exempt From % Ownership';
+  let errorMsg = false;
+  const cells = default_headers.split(',');
+  for (let i = 1; i < cells.length; i++) {
+    const result = header_row.includes(cells[i]);
+    if (!result) {
+      errorMsg = 'The header row is missing fields';
+      break;
+    }
+  }
+  return errorMsg;
+};
+
 export function parseWhitelistCsv(file: string) {
   let data;
-  let parseError = '';
-  try {
-    data = csvParse(file, {
-      skip_empty_lines: true,
-      trim: true,
-      cast: (rawValue, context) => {
-        const value = rawValue;
+  let parseError = checkCSVHeaders(file);
+  if (!parseError) {
+    try {
+      data = csvParse(file, {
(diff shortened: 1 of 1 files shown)


packages/polymath-issuer/src/utils/parsers/index.js
```diff
@@ -83,68 +83,89 @@ const checkColumnCount = data => {
   return errorMsg;
 };
 
+const checkCSVHeaders = data => {
+  // Split the input into lines
+  let rows = data.split('\n');
+  const header_row = rows[0].split(',');
+  const default_headers =
+    'Address,Sale Lockup,Purchase Lockup,KYC/AML Expiry,Can Buy From STO,Exempt From % Ownership';
```
josepmc: Is Exempt from % ownership a default header? The user may not have the max % ownership enabled.

packages/polymath-issuer/src/utils/parsers/index.js
```diff
@@ -83,68 +83,89 @@ const checkColumnCount = data => {
   return errorMsg;
 };
 
+const checkCSVHeaders = data => {
```
josepmc: What's the behaviour with the isAccredited and nonAccreditedLimit fields?

packages/polymath-issuer/src/utils/parsers/index.js
```diff
@@ -83,68 +83,89 @@ const checkColumnCount = data => {
...
+            return false;
+          }
+          if (value.split('/').length === 3) {
+            const [rawMonth, day, rawYear] = value.split('/');
+            // Months are 0-based
+            const month = parseInt(rawMonth, 10) - 1;
+            let year = parseInt(rawYear, 10);
+
+            // Support for incomplete years
+            if (year < 2000) {
```
josepmc: Are we sure we want to have something like this? The issuer inputs 31/12/1999 and then we bump it to 31/12/3999?

packages/polymath-issuer/src/utils/parsers/index.js
sajclarke: Yes it's default. It'll be empty if not enabled.

packages/polymath-issuer/src/utils/parsers/index.js
sajclarke: Good point. The intention is probably to capture if the value is between 1 and 100 then adjust it accordingly. Would be easier to invalidate the year instead I think. Thoughts @RafaelVidaurre ?

packages/polymath-issuer/src/utils/parsers/index.js
sajclarke: I think those fields are optional (if usdsto is attached)

packages/polymath-issuer/src/utils/parsers/index.js
RafaelVidaurre: As discussed on slack, yup, let's fix this.

PR #308, opened 2019-01-18
Feature/support hardware wallets

RafaelVidaurre:

# This PR

Uses personal_sign for signing instead of signTyped which is at this point not supported by metamask / ledger.

### DIFF
--- a/packages/polymath-ui/src/redux/account/actions.js
+++ b/packages/polymath-ui/src/redux/account/actions.js
@@ -65,6 +65,59 @@ export type AccountInnerData = {|
   email: string,
 |};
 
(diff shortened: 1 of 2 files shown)


packages/polymath-ui/src/redux/account/actions.js
```diff
@@ -64,6 +65,59 @@ export type AccountInnerData = {|
   email: string,
 |};
 
+const signTypedData = ({ web3, normal, typed, address }) => {
```
monitz87: Where are you using this function?

RafaelVidaurre: Left it for future reference when we migrate this code. Adding a comment though!

PR #318, opened 2019-01-23
fix(shared): display raised amount to 20 significant digits

sajclarke:


packages/polymath-ui/src/helpers.js
```diff
@@ -49,6 +49,19 @@ export const thousandsDelimiter = (v: number) => {
...
+  decimals: number
+) => {
+  const str = data.toString();
+  if (str.indexOf('e') !== -1) {
+    return Number(data.toPrecision(precision)).toExponential(decimals);
+  } else {
+    return parseFloat(data.toFixed(decimals));
+  }
+};
+
```
Grsmto: @RafaelVidaurre if you could take a look at that as you worked with the numbers. Shouldn't BigNumber lib that we are using handle this?
Also if this function is actually necessary we should have it in `shared` and `new-shared` instead of here, that file is deprecated (might be worth adding a comment in there).

sajclarke: I've made these changes

packages/polymath-ui/src/helpers.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 
 import React from 'react';
+import { BigNumber } from 'bignumber.js';
```
Grsmto:
```suggestion
```

You can remove this line.

packages/polymath-shared/src/utils/format.js
```diff
@@ -87,3 +87,16 @@ export const toTokens = (
   const number = new BigNumber(value);
   return number.toFormat(decimals);
 };
+
+export const toLargeNumber = (
+  data: number,
+  precision: number,
+  decimals: number
+) => {
+  const str = data.toString();
```
RafaelVidaurre: @sajclarke not sure what this method is doing. Can you explain a bit in more detail what it does?

RafaelVidaurre: @sajclarke not sure what this method is doing. Can you explain a bit in more detail what it does?

sajclarke: This method is for formatting the amount raised by STO. It checks if the value is an exponential or not and ensures that it is displayed with (up to) 6 decimals only

PR #323, opened 2019-01-28
fix(issuer): submit ticker symbol uppercase

sajclarke:

### DIFF
--- a/packages/polymath-issuer/src/pages/ticker/components/ReserveTickerForm.js
+++ b/packages/polymath-issuer/src/pages/ticker/components/ReserveTickerForm.js
@@ -127,8 +127,8 @@ const formikEnhancer = withFormik({
   },
   handleSubmit: (values, { props }) => {
     const { dispatch } = props;
-    values.ticker = values.ticker.toUpperCase();
-    dispatch(reserve(values));
+    const formattedValues = { ...values, ticker: values.ticker.toUpperCase() };
+    dispatch(reserve(formattedValues));
   },
 });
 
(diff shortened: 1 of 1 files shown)


packages/polymath-issuer/src/pages/ticker/components/ReserveTickerForm.js
```diff
@@ -60,6 +60,12 @@ export const ReserveTickerFormComponent = ({ handleSubmit }) => (
         <FormItem.Input
           component={TextInput}
           placeholder="Up to 10 characters (example: TORO-A)"
+          onChange={e => {
+            let value = e.target.value || '';
+            value = value.toUpperCase().trim();
+            // setFieldValue(field.name, value);
+            e.target.value = value;
```
Grsmto: I think it's less bug-prone to do this in CSS instead with `text-transform: uppercase` for the display of the input. And do like you do with `.toUpperCase()` on submit.

packages/polymath-issuer/src/pages/ticker/components/ReserveTickerForm.js
```diff
@@ -127,7 +127,7 @@ const formikEnhancer = withFormik({
   },
   handleSubmit: (values, { props }) => {
     const { dispatch } = props;
-
+    values.ticker = values.ticker.toUpperCase();
```
Grsmto: Hey, just a comment here, we should avoid mutating objects like you are doing here. It can lead to unexpected behaviour: for example that's the reason why React has a `setState` function and we do `this.setState({ blabla }) and not `this.state = { blabla }`.
So here instead you should better do something like:
`const formattedValues = Object.assign({}, values, { ticker: ticker.toUpperCase() }` (or
`const formattedValues = { ...values, ticker: ticker.toUpperCase() }`)
This way you don't mutate `values` but create a new object `formattedValues`.

sajclarke: Agreed

Grsmto:
LGTM.

PR #328, opened 2019-01-29
Feature/Checkpoints list presenter

Grsmto:

- [ ] I linked the issues related to this PR in its description (if any).

- Style list of dividends checkpoints
- Button component now supports an `iconPosition` property to simplify usage of buttons that have an icon inside them (similar API as Semantic UI).
- Add Luxon as dependency to manage dates, instead of Moment.js.
- Add new `IconOutlined` component so icons can be circled as an outline instead of plain background.
- Add a `ProgressIndicator` component that can be used horizontally and vertically.


RafaelVidaurre:
@Grsmto I think you merged from `develop` after @monitz87 rolled back to `2.0`. Which means that merging this would also rollback the dividends feature to 2.0 💀

RafaelVidaurre:
@monitz87 since this PR uses `Luxon` and I know you were adding momentjs in some pr, mind using Luzon instead?

PR #329, opened 2019-01-29
fix(issuer): check spent poly balance before asking for faucet

sajclarke:

### DIFF
--- a/packages/polymath-issuer/src/actions/sto.js
+++ b/packages/polymath-issuer/src/actions/sto.js
@@ -40,7 +40,7 @@ export const data = (contract: STO, details: ?STODetails) => ({
 
 const STO_TYPE = 3;
 
-const ConfigSTOConfirmContent = ({ setupCost }) => (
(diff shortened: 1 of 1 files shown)


packages/polymath-issuer/src/actions/sto.js
```diff
@@ -248,11 +249,18 @@ export const configureSTO = (config: STOConfig, type: string) => async (
         const stoModuleConfig = {
           ...config.data,
         };
-
         const tokenBalance = await PolyToken.balanceOf(token.address);
         const titles = ['Deploying And Scheduling'];
 
         if (tokenBalance.lt(setupCost)) {
+          if (!hasEnoughBalance) {
```
RafaelVidaurre: It looks like this shouldn't be required if we are already checking in the above line for `tokenBalance.tl(setupCost)`.

Also, it is safer to get the balance right before we do the action as it is done in line `252`

RafaelVidaurre: Nevermind, this is checking for the Security Token's balance

PR #340, opened 2019-01-31
Feature/transaction queue modal text

monitz87:

This PR adds support for dynamic text in the transaction modals. This means that procedure-specific and transaction-specific parameters can be displayed (and customized) on the modals via special content mapper function props.

**NOTE:** the default mappers are set up and ready to use the transaction and queue arguments, but are returning placeholder text at the moment. I asked Thomas if he (or somebody else from PM) could help me out compiling all the text so as to not go diving in legacy looking for it, and he said he would see what he could do. Until I get an answer from him, the placeholders stay. I'll leave a task on asana to replace the placeholders with real text as a reminder, but this is ready for review/merge either way.


packages/new-polymath-ui/src/components/ModalTransactionQueue/TransactionItem/TransactionItem.tsx
```diff
@@ -78,3 +81,9 @@ export const TransactionItem = ({ transaction }: TransactionItemProps) => {
     </sc.Wrapper>
   );
 };
+
+TransactionItem.defaultProps = {
+  getTitle: getTransactionTitle,
+};
```
Grsmto: so we really need to have the defaultProps defined 2 times? The one inlined in the function declaration is really bloating readability :(

monitz87: The one inlined is not a definition, it's a type declaration. We could probably refactor it so that the type is defined separately for better readability, but it is necessary for it to be typesafe

monitz87: Changed it to be more readable

PR #342, opened 2019-01-31
Feature/dividends list presenter

Grsmto:
- Add `ProgressIndicator` component, supporting vertical/horizontal layout and unordered/ordered
- Rewrite `Tooltip` component using popper.js instead of react-tooltip so we have a much more natural API: 
```
<Button>
  Hover me!
  <Tooltip>This is a tooltip</Tooltip>
</Button>
```
- Add Luxon as Date library (instead of Moment.js)

- Dividends (horizontal) list

screenshot 2019-02-04 at 12 22 56


Grsmto:
Closing because branch was based on `develop` instead of `dividends`. See #347.

PR #343, opened 2019-01-31
Feature/downloadable file generation

monitz87:

This PR:
- Implements a utility that generates a CSV file from JSON data at runtime and downloads it in the browser
- Uses said utility to implement exporting the ownership list of a checkpoint in the checkpoint container

**NOTE:** The only thing that's missing to finish the checkpoint container after merging this PR is to pass the props to the presenter once it is implemented


packages/new-polymath-issuer/package.json
```diff
@@ -58,6 +58,7 @@
     "@polymathnetwork/sdk": "0.0.0",
     "bignumber.js": "8.0.1",
     "lodash": "^4.17.11",
+    "moment": "^2.24.0",
```
RafaelVidaurre: Can we use that other time library @Grsmto added in his PR?

PR #348, opened 2019-02-04
Feature/connect create checkpoint logic

monitz87:

This PR connects the Dividends page container to its presenter and fixes some typing errors

### DIFF
--- a/packages/new-polymath-issuer/package.json
+++ b/packages/new-polymath-issuer/package.json
@@ -85,6 +85,7 @@
     "@types/react-dom": "^16.0.11",
     "@types/react-redux": "^6.0.10",
     "@types/redux-mock-store": "^1.0.0",
+    "@types/styled-components": "^4.1.8",
     "autoprefixer": "^9.4.4",
     "babel-core": "^7.0.0-bridge.0",
     "babel-loader": "8.0.2",
--- a/packages/new-polymath-issuer/src/components/Checkpoints/styles.ts
+++ b/packages/new-polymath-issuer/src/components/Checkpoints/styles.ts
@@ -4,15 +4,27 @@ import {
   List,
   Heading,
   IconOutlined,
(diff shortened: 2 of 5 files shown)


packages/new-polymath-issuer/package.json
```diff
@@ -85,6 +85,7 @@
     "@types/react-dom": "^16.0.11",
     "@types/react-redux": "^6.0.10",
     "@types/redux-mock-store": "^1.0.0",
+    "@types/styled-components": "^4.1.8",
```
Grsmto: I think we should have all the types in the main `package.json` so we avoid version clashes etc.

RafaelVidaurre: The problem is that if an npm package is downloaded from npm (imagine `polymath-ui`) it will not contain all its dependencies specified in the package.json and therefore it will fail.

This means that we must list *all* dependencies that each individual package requires.

Grsmto: If we support npm for individual packages, then it means they need to be transpiled and not expose raw TS. In that case if they are transpiled, the types dependencies are not needed for the package to be used.

RafaelVidaurre: @Grsmto not the type dependencies but what about the content of node_modules?

And yes, they will be transpiled on publish as they should be, this doesn't mean that we are bundling in there its dependencies I believe though

RafaelVidaurre: Oh wait @Grsmto I didn't realize we were talking about type dependencies specifically, but @types depend on the dependency every individual package has though right? So if they match then they will be hoisted and shared, but otherwise they are package-specific... I'm not fully convinced I'm right though, adding a Tech-debt task in Asana to make a decision. I'd tackle this separately from this PR though, just move all type deps everywhere to the root package.json

PR #359, opened 2019-02-07
Fix/race condition on get network

RafaelVidaurre:


packages/new-polymath-sdk/src/Polymath.ts
```diff
@@ -90,11 +73,27 @@ export class Polymath {
...
+      this.lowLevel = new LowLevel({ provider: this.httpProviderUrl });
+    } else {
+      this.lowLevel = new LowLevel();
+    }
+
+    this.polymathRegistryAddress = polymathRegistryAddress;
 
-    this.networkId = await lowLevel.getNetworkId();
-    const account = await lowLevel.getAccount();
+    this.networkId = await this.lowLevel.getNetworkId();
```
monitz87: Consider assigning lowLevel to a variable to avoid repetition of `this`

PR #360, opened 2019-02-07
Feature/dividends  plug into legacy

RafaelVidaurre:

# This PR:
- Adds the required configuration to allow our two methods for incorporating Mordern code into Legacy
- Plugs in Dividends into Legacy

### DIFF
--- a/packages/polymath-issuer/src/index.js
+++ b/packages/polymath-issuer/src/index.js
@@ -1,48 +1,100 @@
-// @flow
+/**
+ * This code is here to enable the progressive migration from Legacy to Modern.
+ * There are several levels in which Modern can replace Legacy functionality
+ * for a given route:
+ * 1. App level: Runs Modern on its own, Legacy is not used for this route
+ * 2. Router level: A subtree of Legacy is replaced by a subtree imported
+ * from Modern.
+ *
+ * There are two main capabilities that this setup allows:
+ * 1. To render the Modern or Legacy dApp depending on the current route.
+ * 2. To transform routes before mounting either app.
+ *
+ * NOTE: In theory we could also have logic to resolve values required for the
+ * redirected route that might not exist yet. Though this would require, an
+ * extra initialization step that might complicate things, so if possible lets
+ * avoid it.
+ */
 
-// Import Sass styles. Order is important.
-// $FlowFixMe
-import './style.scss';
+const redirectors = [
+  /*
+  Example:
 
-import React, { Fragment } from 'react';
-import { render } from 'react-dom';
-import { Provider } from 'react-redux';
-import { ConnectedRouter } from 'connected-react-router';
-import { ThemeProvider } from 'styled-components';
-import { theme, GlobalStyles } from '@polymathnetwork/ui';
-import * as Sentry from '@sentry/browser';
+  {
+    pattern: /dashboard\/(.+)\/dividends/,
+    makeRoute: (params: string[]) => {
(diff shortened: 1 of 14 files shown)


packages/polymath-ui/src/components/EthNetworkWrapper/actions.js
```diff
@@ -170,6 +171,11 @@ export const init = (networks: Array<string>) => async (dispatch: Function) => {
     web3WS,
   };
 
+  await polyClient.connect({
```
monitz87: Aren't the modern sagas handling this?

packages/polymath-issuer/src/index.js
```diff
@@ -1,48 +1,101 @@
...
-import routes from './routes';
-
-unregister();
-
-if (process.env.NODE_ENV === 'production') {
-  // Init error monitoring tool
-  Sentry.init({
-    dsn: SENTRY_KEY,
-    environment: process.env.REACT_APP_DEPLOYMENT_STAGE,
+import { find } from 'lodash';
```
Grsmto: Useless import here.

packages/polymath-issuer/src/index.js
```diff
@@ -1,48 +1,101 @@
...
+  modernAppRoutePatterns.forEach(pattern => {
+    if (match) {
+      return;
+    }
+    match = path.match(pattern);
+  });
+
+  if (match) {
+    console.log('Mounting: Modern dApp');
+    require('./mountModern');
```
Grsmto: Maybe you could use Webpack code-splitting instead here:
`import('./mountModern')`
This way when you load the app the first time, you would actually load only 1 app. That might not change much because we have zero performances optimisation but as the current bundle is 8mb, you would load 8mb * 2 of JS bundle if you load the 2 apps, might start to be a little big 😄

RafaelVidaurre: Yeah we could do that. I figured it was not worth the effort given this is only going to live for ~2 months

packages/polymath-ui/src/components/EthNetworkWrapper/actions.js
RafaelVidaurre: Not for this scenario. The sagas do this based on Modern's routing which doesn't exist for when replacing components.

packages/polymath-issuer/src/index.js
```diff
@@ -1,48 +1,101 @@
...
+
+  // If there's no redirection we render the proper app
+  if (!redirectionMatch) {
+    mountApp();
+    return;
+  }
+
+  // Parse parameters from redirector's regex
+  const params = redirectionMatch.slice(1);
+  const getTargetRoute = matchingRedirector.makeRoute(params);
```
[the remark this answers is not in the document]
RafaelVidaurre: Whoops, yep. Good catch

packages/polymath-issuer/src/index.js
RafaelVidaurre: Thanks Adrien 👍 !

packages/polymath-issuer/src/index.js
RafaelVidaurre: If this is a quick win we should do it though. TBH I've never had to use dynamic imports before, you think this would be easy?

RafaelVidaurre: Well, let's tackle this in a separate PR if we want to so that we don't postpone this merge any further. I know @josepmc is impatient

PR #365, opened 2019-02-12
fix(shared): refresh countdown timer when it reaches 0

sajclarke:


packages/polymath-ui/src/components/STOStatus/STOStatus.js
```diff
@@ -133,7 +178,7 @@ export default class STOStatus extends Component<Props> {
               </div>
             </div>
           </div>
-          {countdownProps != null && (
+          {countdownProps != null && !isSaleClosed ? (
```
Grsmto: `isSaleClosed` doesn't actually get set to `true` anywhere. So it will always be `false` anyway. I think something is missing here.

sajclarke: That's correct. The `isSaleClosed` is no longer necessary so I've removed it

PR #371, opened 2019-02-15
fix(issuer): special characters validation

BilalAlGhazi:

Validate the following condition of the ticker:
- Ticker must start with a letter or number (and not with special characters)
- Ticker must end with a letter or number (and not with special characters)
- Ticker must not contain 2 special characters together

### DIFF
--- a/packages/polymath-issuer/src/pages/ticker/components/ReserveTickerForm.js
+++ b/packages/polymath-issuer/src/pages/ticker/components/ReserveTickerForm.js
@@ -21,11 +21,11 @@ const formSchema = validator.object().shape({
     .isRequired(requiredMessage)
     .max(10, maxMessage)
     .matches(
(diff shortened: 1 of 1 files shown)


packages/polymath-issuer/src/pages/ticker/components/ReserveTickerForm.js
```diff
@@ -24,6 +24,12 @@ const formSchema = validator.object().shape({
       /^[a-z0-9./-]+$/i,
```
monitz87:
```suggestion
      /^[a-z0-9.-]+$/i,
```

Leftovers from boris' reign of terror

RafaelVidaurre: Thanks @monitz87 didn't catch those

PR #382, opened 2019-02-21
feature/csv parser enhancer

BilalAlGhazi:

## This PR Includes:
Enhancer component to provide CSV parsing functionality as render-props.

### DIFF
--- a/packages/new-polymath-issuer/src/components/enhancers/ParseCsv.tsx
+++ b/packages/new-polymath-issuer/src/components/enhancers/ParseCsv.tsx
@@ -48,23 +48,23 @@ export class ParseCsv extends Component<Props, State> {
   };
 
   public setFile = async (file: string | File) => {
+    const {
+      missingRequiredColumnsErrorMessage,
+      extraColumnsErrorMessage,
+      rowsExceedMaxLimitErrorMessage,
+      parseErrorMessage,
+      ...csvConfig
+    } = this.props.config;
+    const fileProps: csvParser.Props = {
+      data: file,
+      ...csvConfig,
+      errorMessages: {
+        missingRequiredColumns: missingRequiredColumnsErrorMessage || '',
+        extraColumns: extraColumnsErrorMessage || '',
+        rowsExceedMaxLimit: rowsExceedMaxLimitErrorMessage || '',
+      },
+    };
     try {
-      const {
-        missingRequiredColumnsErrorMessage,
-        extraColumnsErrorMessage,
-        rowsExceedMaxLimitErrorMessage,
-        parseErrorMessage,
-        ...csvConfig
-      } = this.props.config;
-      const fileProps: csvParser.Props = {
-        data: file,
-        ...csvConfig,
-        errorMessages: {
-          missingRequiredColumns: missingRequiredColumnsErrorMessage || '',
-          extraColumns: extraColumnsErrorMessage || '',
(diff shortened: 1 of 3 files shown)


packages/new-polymath-issuer/src/components/enhancers/ParseCsv.tsx
```diff
@@ -0,0 +1,77 @@
...
+      validRows: 0,
+      errorRows: 0,
+      ignoredRows: 0,
+      isFileValid: false,
+      errors: [],
+    },
+  };
+
+  public setFile = (fileProps: csvParser.Props) => {
+    csvParser
```
RafaelVidaurre: We prefer using async/await here.

Something like:

```ts
public setFile = async (fileProps: csvParser.Props) => {
  try {
    const parseResult = await csvParser.parseCsv(fileProps);
    this.setState({ data: parseResult });
  } catch {
    this.setState(/ * ... */);
  }
}
```

packages/new-polymath-issuer/src/components/enhancers/index.tsx
```diff
@@ -1,2 +1,3 @@
 export { DataFetcher } from './DataFetcher';
 export { PageLoader } from './isPage';
+export { ParseCsv, RenderProps } from './ParseCsv';
```
RafaelVidaurre: I think you don't need to export the `RenderProps` type. The enhancer should provide these out of the box

packages/new-polymath-issuer/src/components/enhancers/__tests__/ParseCsv.tsx
```diff
@@ -0,0 +1,106 @@
...
+      setFile = inputParams.setFile;
+      clearFile = inputParams.clearFile;
+      result = inputParams.data.result;
+      return <div>Hello</div>;
+    };
+    render(
+      <ParseCsv parseErrorMessage="Parsing Error" render={renderFunction} />
+    );
+
+    setFile({
```
RafaelVidaurre: I think there was a confusion with how I suggested the column data was done. What I mean was that, instead of passing the configuration here, I suggested we did it in the component itself:

```ts
const fileConfig = {
  data: `column 1, column 2`
/* ... etc ... */
};

render (
  <ParseCsv config={fileConfig} {/* ... */} />
)
```

If we don't do this then our Presenter becomes aware of the file structure, which is something we shouldn't enforce

packages/new-polymath-issuer/src/components/enhancers/__tests__/ParseCsv.tsx
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
+import { render, cleanup, wait } from 'react-testing-library';
```
RafaelVidaurre: No need to import `cleanup`. It is being called after each test on our global setup

packages/new-polymath-issuer/src/components/enhancers/ParseCsv.tsx
```diff
@@ -0,0 +1,103 @@
...
+      validRows: 0,
+      errorRows: 0,
+      ignoredRows: 0,
+      isFileValid: false,
+      errors: [],
+    },
+  };
+
+  public setFile = async (file: string | File) => {
+    try {
```
RafaelVidaurre: Instead of wrapping the whole method in a try catch, I'd rather only wrap the part that is actually being handled, which is the line with the `await` call. This is so that we don't mix up unrelated errors in that handler

PR #390, opened 2019-02-26
Feature/support 2 0 0 capped sto

RafaelVidaurre:

### DIFF
--- a/packages/polymath-js/src/contracts/STO.js
+++ b/packages/polymath-js/src/contracts/STO.js
@@ -1,6 +1,7 @@
 // @flow
 
 import artifact from '@polymathnetwork/polymath-scripts/fixtures/contracts/CappedSTO.json';
+import IModuleFactoryArtifacts from '@polymathnetwork/polymath-scripts/fixtures/contracts/IModuleFactory.json';
 import BigNumber from 'bignumber.js';
 
 import Contract from './Contract';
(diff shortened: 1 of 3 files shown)


packages/polymath-js/src/contracts/STO.js
```diff
@@ -74,7 +95,7 @@ export default class STO extends Contract {
       tokensSold: this.token.removeDecimals(tokensSold),
       capReached,
       isOpen,
-      rate,
+      rate: rate,
```
RafaelVidaurre: Whoops

packages/polymath-js/src/contracts/STO.js
```diff
@@ -48,13 +49,33 @@ export default class STO extends Contract {
...
+    // with 0.0.0 version of the Capped STO. This will be tackled properly by
+    // module version supporton the SDK
+    const version = (await GenericModuleFactory.methods
+      .getUpperSTVersionBounds()
+      .call()).join('.');
+
+    let rate = rawRate;
+
+    if (version === '0.0.0') {
+      rate = this._fromWei(rate).toNumber();
```
monitz87: Shouldn't this comparison be `!==`?

packages/polymath-js/src/contracts/STO.js
```diff
@@ -48,13 +49,33 @@ export default class STO extends Contract {
...
+    const GenericModuleFactory = new Contract._params.web3.eth.Contract(
+      IModuleFactoryArtifacts.abi,
+      factoryAddress
+    );
+
+    // NOTE @RafaelVidaurre: This is a hacky implementation to be backwards-compatible
+    // with 0.0.0 version of the Capped STO. This will be tackled properly by
+    // module version supporton the SDK
+    const version = (await GenericModuleFactory.methods
+      .getUpperSTVersionBounds()
```
monitz87: This should be a call to `version`

packages/polymath-js/src/contracts/STO.js
RafaelVidaurre: Nope. version `0.0.0` is the one existing in `2.0.0` contracts

packages/polymath-js/src/contracts/STO.js
RafaelVidaurre: @monitz87 `version` doesn't exist in `0.0.0`. All I couldn found was `getUpperSTVersionBounds`

monitz87: `getUpperSTVersionBounds` gets the highest security token version that supports that module, it doesn't get the module version. Also, I'm looking at 2.0 `ModuleFactory.sol` and I see `string public version` defined

packages/polymath-js/src/contracts/STO.js
monitz87: What I mean is that in 2.0 the rate **doesn't** come as wei, so you should only make the conversion if the version is 2.1

RafaelVidaurre: What, really?

Talk to you on Slack, I'm not sure this is the case

packages/polymath-js/src/contracts/STO.js
RafaelVidaurre: @monitz87 will give it another look, I looked for the same thing since I share your concerns with `getUpperSTVersionBounds`

PR #397, opened 2019-02-27
Feature/dividends wizard part 3

BilalAlGhazi:

## This PR:

Contains the `Presenter` for step 3 of the Dividends Wizard.
screen shot 2019-02-27 at 6 24 33 pm


packages/new-polymath-ui/src/components/FormItem/Input/index.tsx
```diff
@@ -8,6 +8,8 @@ export interface Props
   FormikComponent: React.ComponentType<any>;
   placeholder?: string;
   component: React.ComponentType<FormikExternalProps>;
+  unit?: string;
+  label?: string;
```
Grsmto: @RafaelVidaurre here we face an issue with the typings: as this component takes a `component` prop that can be any possible component, and forward the props, it means TS complains if you apply any prop that is not in the interface of `FormItem.Input` itself. ex: `<FormItem.Input component={Select} someObscureSelectProp="fu" />` it will say that `someObscureSelectProp` is missing from `FormItem.Input` interface.
Here Bilal added the props to the interface but that's not scalable...Do you know a solution for this?

RafaelVidaurre: I'll have a look, though I have my suspicions as to why this could be. Will get back to you in a bit

PR #403, opened 2019-03-01
Fix/edit dividends module address

RafaelVidaurre:

### DIFF
--- a/packages/new-polymath-ui/src/validator/index.ts
+++ b/packages/new-polymath-ui/src/validator/index.ts
@@ -1,19 +1,45 @@
 import * as Yup from 'yup';
 import { isValidAddress } from '@polymathnetwork/sdk';
 import { BigNumberSchema } from './schemas/BigNumberSchema';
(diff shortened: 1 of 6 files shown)


packages/new-polymath-ui/src/validator/index.ts
```diff
@@ -1,19 +1,45 @@
...
-const validator: typeof Yup & { bigNumber: () => BigNumberSchema } = {
+type ValidatorFn<T> = (
+  message?:
+    | string
+    | ((params: object & Partial<Yup.TestMessageParams>) => string)
+    | undefined
+) => T;
+
+interface CustomStringSchema extends Yup.StringSchema {
+  isAddress: ValidatorFn<this>;
```
Grsmto: `isAddress` could be more specific. It could refer to other things in this context like a postal address for example. I think `isWalletAddress` would fit better.

RafaelVidaurre: Makes sense, making the change! Calling it `isEthereumAddress` since that's what it is checking for

PR #408, opened 2019-03-04
Fix/etherscan url per environment

RafaelVidaurre:


packages/new-polymath-issuer/src/components/ModalTransactionQueue/Container.tsx
```diff
@@ -69,15 +70,20 @@ export class ModalTransactionQueueContainerBase extends Component<Props> {
   };
 
   public render() {
-    const { transactionQueue } = this.props;
+    const { transactionQueue, networkId } = this.props;
     const { onContinue, onConfirm, onClose } = this;
 
+    const transactionLinkSubdomain = networkId
```
monitz87: This is just me nitpicking, but why do you use `''` here and `undefined` in `packages/new-polymath-issuer/src/pages/SecurityTokensDividends/Container.tsx`? Also the check you're doing is different in both

RafaelVidaurre: that's fine, I'll see if I can improve this. The check I did in TS is different because of typing if I remember correctly. But fair enough, I can be more consistent for ;p

PR #416, opened 2019-03-06
Fix/dividends style fixes

Grsmto:

- missing legacy layout integration


packages/new-polymath-issuer/src/components/DividendList/Presenter.tsx
```diff
@@ -52,12 +54,17 @@ export const DividendListPresenter = ({
           </sc.NewDividendButton>
         </Fragment>
       ) : (
-        <sc.PlaceholderButton href={newDividendUrl}>
+        <sc.PlaceholderButton
+          RouterLink={Link}
```
RafaelVidaurre: `RouterLink` is going to default to the right `Link` if left out. (react-router-dom's or redux-little-router's depending on the dApp)

RafaelVidaurre: (So we should be able to remove these)

packages/new-polymath-ui/src/components/Table/Rows/Row/index.tsx
```diff
@@ -12,15 +12,15 @@ export const Row: FC<Props> = ({ row, small }) => {
     <sc.Row
       {...row.getRowProps()}
       selected={row.isSelected}
-      hasError={!row.isValid}
+      hasError={row.isValid === false}
```
RafaelVidaurre: Why was this change required?

Grsmto: Because if `isValid` is not defined it shouldn't count has being `false`. The `isValid` property can or not be defined based on what "feature" is activated in the `Table` (here it's the `useCsvParser` feature).

RafaelVidaurre: @Grsmto but `undefined` is a falsy value. Therefore you can just do `!row.isValid` which by javascript's type coercion will become a boolean

RafaelVidaurre: Oooh, wait so we have 3 states? valid, invalid and neither?

PR #425, opened 2019-03-06
Fix display of cappedsto raisedamount

sajclarke:
This PR fixes the removal of trailing zeros and resizes the raised amount container since the max digits for hard cap is now set

### DIFF
--- a/packages/polymath-shared/src/utils/format.js
+++ b/packages/polymath-shared/src/utils/format.js
@@ -95,5 +95,5 @@ export const toTokens = (
   }
   return new BigNumber(number.toPrecision(precision))
     .toFormat(decimals, 1)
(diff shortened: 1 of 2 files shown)


packages/polymath-shared/src/utils/format.js
```diff
@@ -95,5 +95,5 @@ export const toTokens = (
   }
   return new BigNumber(number.toPrecision(precision))
     .toFormat(decimals, 1)
-    .replace(/([0-9]+(\.[0-9]+[1-9])?)(\.?0+$)/, '$1');
+    .replace(/\.?0+$/, '');
```
RafaelVidaurre: This would remove all zeroes on the number.

PR #433, opened 2019-03-07
Feature/dividends details presenter data

monitz87:

This PR completes the dividends details screen functionality and fixes the calculations and display data


packages/new-polymath-issuer/src/pages/Home.tsx
```diff
@@ -90,12 +92,24 @@ export class ContainerBase extends Component<Props> {
         name: 'My Dividend Distribution',
         amount: new BigNumber('10000'),
         checkpointId: 1,
-        excludedAddresses: [],
+        excludedAddresses: ['0x821aea9a577a9b44299b9c15c88cf3087f3b5544'],
+        // excludedAddresses: [],
```
RafaelVidaurre: 👁

packages/new-polymath-issuer/src/pages/Home.tsx
```diff
@@ -90,12 +92,24 @@ export class ContainerBase extends Component<Props> {
...
+      updateTaxWithholdingListStart({
+        securityTokenSymbol: 'A0T0',
+        dividendType: DividendModuleTypes.Erc20,
+        investorAddresses: ['0x0d1d4e623d10f9fba5db95830f7d3839406c6af2'],
+        percentages: [0.5],
+      })
+    );
+  };
+
   public checkForValidity = async () => {
```
RafaelVidaurre: This log is from before this PR apparently, but lets remove it

packages/new-polymath-issuer/src/pages/Home.tsx
monitz87: Remember that this is the home screen we use to test transactions. It's not really production code so I don't see the need to keep it clean

packages/new-polymath-issuer/src/pages/Home.tsx
monitz87: Same as above

packages/new-polymath-issuer/src/pages/Home.tsx
RafaelVidaurre: Oh right, missed that. Nvm then!

PR #437, opened 2019-03-08
fix(new-issuer): dividends wizard step 2 validation

BilalAlGhazi:

## This PR:

Fixes the bug (No input validation on edit a tax withholding investor)
https://app.asana.com/0/951808452757493/1113255342169342


RafaelVidaurre:
Approved, though there's two objects typed as `any` in the validation callback. I understand we don't benefit from those types in this scenario, but aren't there types we can use for these?

PR #438, opened 2019-03-08
Feature/batch withholding list transactions

monitz87:

This PR modifies the tax withholding list procedure to use multiple transactions (one for each 200 addresses). It also changes some parameter names for consistency


packages/new-polymath-ui/src/components/utils/contentMappings.ts
```diff
@@ -609,27 +622,11 @@ export const getTransactionContent = (
       // TODO @monitz87: deal with this after we decide how to handle "batching" multiple transactions into
```
RafaelVidaurre: Is this comment still relevant?

monitz87: Yes, it should be changed though

PR #447, opened 2019-03-11
feat(new-issuer): add step3 validations

RafaelVidaurre:


RafaelVidaurre:
@monitz87 nopes, formik does not do that on its own. If you want the form to behave as an actual form it needs to have a form tag. (Formik even exports their own `Form` component which auto-binds the `onSubmit`)

PR #448, opened 2019-03-11
Feature/csv duplicate checking

BilalAlGhazi:

## This PR
Adds a generic validation function to the `CsvUploader` component. This function can be used to handle duplicate records.


packages/new-polymath-shared/src/utils/csvParser.ts
```diff
@@ -1,11 +1,13 @@
 import _ from 'lodash';
 import Papa, { ParseResult } from 'papaparse';
 import * as validators from './validators';
+import { string } from 'yup';
```
RafaelVidaurre: Not sure why we are importing this here

BilalAlGhazi: I think the editor imported it for some reason.

PR #452, opened 2019-03-11
Feature/dividend wizard summary box data

monitz87:

This PR:
- Adds an `onChange` function prop to `FormItem.Input` that allows us to hook custom behavior to changes in the input value
- Adds an `onFieldChange`  function prop to the formik `Form` wrapper that allows us to hook custom behavior to changes in any input value
- Invalidates the Tax Withholding cache whenever a new checkpoint is created
- Plugs in the dividend wizard summary box to real data
- Fixes miscellaneous presentation logic bugs

### DIFF
--- a/packages/new-polymath-issuer/src/pages/DividendsWizard/Presenter.tsx
+++ b/packages/new-polymath-issuer/src/pages/DividendsWizard/Presenter.tsx
@@ -20,7 +20,9 @@ import * as sc from './styles';
 import { Step1 } from './Step-1';
 import { Step2 } from './Step-2';
 import { Step3 } from './Step-3';
-import { types } from '@polymathnetwork/new-shared';
+import { types, formatters } from '@polymathnetwork/new-shared';
+import BigNumber from 'bignumber.js';
+import { difference, intersection } from 'lodash';
 
 export interface ExclusionEntry {
   ['Investor ETH Address']: string;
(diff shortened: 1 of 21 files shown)


packages/new-polymath-issuer/src/pages/DividendsWizard/Presenter.tsx
```diff
@@ -50,17 +52,47 @@ export interface Props {
 
 export interface State {
   excludedWallets: null | ExclusionEntry[];
+  dividendAmount: BigNumber;
+  positiveWithholdingAmount: number;
 }
 
 export class Presenter extends Component<Props> {
-  public state = {
+  public state: State = {
```
RafaelVidaurre: The state type should go on `Component<Props, State>`

packages/new-polymath-issuer/src/pages/DividendsWizard/Presenter.tsx
```diff
@@ -186,19 +237,29 @@ export class Presenter extends Component<Props> {
...
+                      </Paragraph>
+                    </Flex>
+                  )}
                 </List>
               </Box>
               <Text as="strong" mt="m">
                 Total Dividend Distribution
               </Text>
               <br />
               {/* NOTE: Mock data */}
```
RafaelVidaurre: We can remove this note now

packages/new-polymath-issuer/src/pages/DividendsWizard/Presenter.tsx
monitz87: Yeah, but it's still needed there, otherwise I get type errors when accessing some properties that are initialized as `null`

RafaelVidaurre: Should still be added to the `Component` generic as a second argument though right?

monitz87: Yeah, I did that

PR #468, opened 2019-03-12
Fix/dividends transation modal scroll

Grsmto:

- Modal now supports global window scroll when passed prop `isCentered={false}`.
- Modal now locks the body scroll when opening/closing (otherwise there would be a double scrollbar on the global mode. It was also required for accessibility anyway).

### DIFF
--- a/packages/new-polymath-ui/src/components/Modal/Modal.tsx
+++ b/packages/new-polymath-ui/src/components/Modal/Modal.tsx
@@ -1,8 +1,7 @@
-import React, { Component } from 'react';
+import React, { FC, useState, useLayoutEffect } from 'react';
 import ReactModal from 'react-modal';
-
 import { withTheme, ThemeInterface, styled } from '~/styles';
-
+import { useLockScroll } from '~/components/utils/hooks';
 import { Header } from './Header';
 import { Body } from './Body';
 import { Footer } from './Footer';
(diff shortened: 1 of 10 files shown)


packages/new-polymath-ui/src/components/Modal/Modal.tsx
```diff
@@ -21,82 +20,72 @@ interface Props {
...
+        afterOpen: 'pui-modal--after-open',
+        beforeClose: 'pui-modal--before-close',
+      }}
+      overlayClassName={{
+        base: `pui-modal__overlay ${className}`,
+        afterOpen: 'pui-modal__overlay--after-open',
+        beforeClose: 'pui-modal__overlay--before-close',
+      }}
+      onRequestClose={handleCloseRequest}
+      // overlayRef={node => (overlayRef = node)}
```
RafaelVidaurre: Let's remove these comments

RafaelVidaurre:
```suggestion
```

packages/new-polymath-ui/src/components/Modal/Modal.tsx
```diff
@@ -21,82 +20,72 @@ interface Props {
...
+      // contentRef={node => (contentRef = node)}
```
RafaelVidaurre:
```suggestion
```

PR #482, opened 2019-03-14
Fix/currency list

BilalAlGhazi:

## This PR:
Fixes the bug: [Displayed currencies in dividends step 3 is only displaying 'POLY', 'DAI' and 'Other ERC20'](https://app.asana.com/0/951808452757493/1113838682012039)


RafaelVidaurre:
About this PR. What I was asking on the standup: These tokens need to be somehow associated to addresses if we want them to work. Is there anything else that needs to be done on the container's side to support this @monitz87 ?

monitz87:
Yeah, their addresses have to be added to `packages/new-polymath-shared/src/constants/network.ts`

BilalAlGhazi:
Changed the PR to WIP till we get confirmation of addresses and get addresses for kovan and local.

PR #485, opened 2019-03-14
Add search to dividend listing

sajclarke:

 ### This PR
Fixes this issue: ['Search' functionality on Dividends Checkpoints page](https://app.asana.com/0/958842839304329/1113333549752153)


packages/new-polymath-issuer/src/components/DividendList/Container.tsx
```diff
@@ -38,6 +42,7 @@ export class DividendListContainerBase extends Component<Props> {
             <DividendListPresenter
               securityTokenSymbol={securityTokenSymbol}
               dividends={sortedDividends}
+              filterDividends={filterDividends}
```
Grsmto: Here instead of adding and extra prop, you could just filter them in here and pass them already filtered here `dividends={sortedDividends}`. It's the container so it can do the filtering job! Presenter just receives the dividends already filtered and shows them.

packages/new-polymath-issuer/src/pages/SecurityTokensDividends/Presenter.tsx
```diff
@@ -197,10 +213,34 @@ export const Presenter: FC<Props> = ({
...
+        <GridRow.Col gridSpan={4}>
+          <FormWrapper<Values>
+            // enableReinitialize
+            initialValues={{
+              searchName,
+            }}
+            // onChange={handleSearchChange}
+            render={({ values }) => (
+              <FormItem name="searchName">
+                <FormItem.Label>Search Text</FormItem.Label>
```
Grsmto: You don't have to use Formik for this. It's a single input field, so just use the `BaseInput` directly.

packages/new-polymath-issuer/src/components/DividendList/Container.tsx
sajclarke: Thanks. I chose to use context Provider instead so in order to use hooks, I kept the logic in the Presenter

Grsmto: So I don't think using a context is needed here. 
What I was suggesting is that you pass the `dividends` directly filtered. No extra prop `filterDividends`.
You can see that the `dividends` prop is already being sorted/filtered, the variable is called `sortedDividends` actually!
So instead of only being sorted, they could be sorted + filtered in the same place.
So no extra prop, context etc.

PR #488, opened 2019-03-15
Fix/dividends token contract decimals

monitz87:

This PR adds support for dividend distributions in tokens with a decimal amount other than 18

### DIFF
--- a/packages/new-polymath-sdk/src/LowLevel/Erc20.ts
+++ b/packages/new-polymath-sdk/src/LowLevel/Erc20.ts
@@ -8,17 +8,17 @@ import {
   AllowanceArgs,
   BalanceOfArgs,
 } from './types';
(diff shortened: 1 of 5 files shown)


packages/new-polymath-sdk/src/LowLevel/Erc20.ts
```diff
@@ -62,16 +72,39 @@ export class Erc20 extends Contract<Erc20Contract> {
...
-    return fromWei(allowance);
+    const decimals = await this.decimals();
+
+    return fromDivisible(allowance, decimals);
+  };
+
+  public decimals = async () => {
+    const { decimalPlaces } = this;
+
+    if (decimalPlaces) {
```
RafaelVidaurre: Should explicitly do a null check

PR #489, opened 2019-03-15
chore(new-ui): modify tooltipicon hover and font size

sajclarke:

### DIFF
--- a/packages/new-polymath-ui/src/components/TooltipIcon/styles.ts
+++ b/packages/new-polymath-ui/src/components/TooltipIcon/styles.ts
@@ -8,10 +8,10 @@ import { ReactNode } from 'react';
 
 export const Icon = styled(IconButton)`
   transition: ${({ theme }) => `color ${theme.transitions.hover.ms}ms`};
-  color: ${({ theme }) => theme.colors.secondary};
+  color: ${({ theme }) => theme.colors.blue[1]};
 
   &:hover {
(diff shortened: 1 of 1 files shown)


packages/new-polymath-ui/src/components/TooltipIcon/styles.ts
```diff
@@ -30,5 +30,5 @@ export const Tooltip = styled(TooltipComponent)<Props>`
   border: 1px solid #dfe3e6;
   word-wrap: break-word;
   font-weight: normal;
-  font-size: ${({ theme }) => theme.fontSizes.baseText};
+  font-size: ${({ theme }) => theme.fontSizes[2]};
```
Grsmto: I think the font size was right already, if you check the Invision. From what I see was reported by Stan, it's the `line-height` that is wrong.

sajclarke: Thanks. Resolved.

Grsmto: Sorry to be annoying again, but could you use the value from the theme for that instead of hardcoded value? 😅
I believe `theme.lineHeights.normal` should work fine!
As a general rule, try to hardcode only if it's not in theme (or if the theme doesn't fit!).

sajclarke: No problem

PR #501, opened 2019-03-18
Fix/100 tax withholding

monitz87:

This PR adds validation to the CSV upload and tax withholding modal to prevent issuers from setting a percentage to 100. This is a temporary fix until the real problem is fixed in the contracts. It also handles some language issues and fixes a DataFetcher error, along with some typing improvements

### DIFF
--- a/packages/new-polymath-issuer/src/pages/DividendsWizard/Step-2/CsvModal.tsx
+++ b/packages/new-polymath-issuer/src/pages/DividendsWizard/Step-2/CsvModal.tsx
@@ -16,7 +16,6 @@ import {
   TaxWithholdingsItem,
   TaxWithholdingStatuses,
 } from './shared';
-import { boolean } from 'yup';
 
 interface Props {
   isOpen: boolean;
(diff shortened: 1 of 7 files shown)


packages/new-polymath-issuer/src/pages/DividendsWizard/Step-2/CsvModal.tsx
```diff
@@ -130,7 +129,12 @@ export const CsvModal: FC<Props> = ({
               },
               {
                 name: csvTaxWithholdingKey,
-                validators: [validators.isNotEmpty],
+                validators: [
+                  validators.isNotEmpty,
+                  validators.isNumber,
+                  validators.isLowerThan(99),
```
RafaelVidaurre: Shouldn't this be `lowerThanOrEqual`?

monitz87: True, I was messing around with those validations and forgot to change that one

PR #508, opened 2019-03-19
fix(new-issuer): prevent 0 amount dividends

BilalAlGhazi:

## This PR:

Fixes the bug [UI allows me to proceed with an amount = 0 when creating a dividend](https://app.asana.com/0/951808452757493/1114514932327309)

### DIFF
--- a/packages/new-polymath-issuer/src/pages/DividendsWizard/Step-3/index.tsx
+++ b/packages/new-polymath-issuer/src/pages/DividendsWizard/Step-3/index.tsx
@@ -25,12 +25,18 @@ import { CreateDividendDistributionParams } from '~/pages/DividendsWizard/Contai
 import { RootState } from '~/state/store';
 import { getApp, getSession } from '~/state/selectors';
 import { connect } from 'react-redux';
-import { validateYupSchema, yupToFormErrors, FormikErrors } from 'formik';
+import {
+  validateYupSchema,
+  yupToFormErrors,
+  FormikErrors,
+  FormikTouched,
+} from 'formik';
 import {
   Wallet,
   GetErc20BalanceByAddressAndWalletArgs,
   GetIsValidErc20ByAddressArgs,
 } from '~/types';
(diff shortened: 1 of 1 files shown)


packages/new-polymath-issuer/src/pages/DividendsWizard/Step-3/index.tsx
```diff
@@ -255,18 +255,43 @@ const Step3Base: FC<Props> = ({
...
+                      key !== 'tokenAddress' ||
+                      currency === types.Tokens.Erc20
+                    ) {
+                      setFieldTouched(`${key}`, true);
+                    }
+                  }
+                }
+                if (
+                  values.dividendAmount &&
+                  values.dividendAmount.isEqualTo(new BigNumber(0))
```
Grsmto: Why do `new BigNumber(0)` instead of just `=== 0`?

BilalAlGhazi: Doing `=== 0` gives an error that the statement will always resolve to `false`:

Screen Shot 2019-03-19 at 8 04 28 AM

BilalAlGhazi: Changed to Formik validation

packages/new-polymath-issuer/src/pages/DividendsWizard/Step-3/index.tsx
```diff
@@ -255,18 +255,36 @@ const Step3Base: FC<Props> = ({
...
+          setFieldTouched,
+          isValid,
+          initialValues,
+          touched,
+        }) => {
           const { currency } = values;
-
           return (
             <form
               onSubmit={submitEvent => {
```
RafaelVidaurre: Lets move this to a handler function instead of having this inline. If you need some of the values in the render context pass them as arguments

PR #517, opened 2019-03-19
Fix/excluded addresses in tax withholdings

RafaelVidaurre:

### DIFF
--- a/packages/new-polymath-issuer/src/pages/DividendsWizard/Step-2/TaxWithholdingModal.tsx
+++ b/packages/new-polymath-issuer/src/pages/DividendsWizard/Step-2/TaxWithholdingModal.tsx
@@ -1,12 +1,6 @@
-import React, { FC, useState, Fragment, useEffect } from 'react';
-import { types } from '@polymathnetwork/new-shared';
-import { get, some, findIndex } from 'lodash';
-import {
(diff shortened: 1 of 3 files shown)


packages/new-polymath-issuer/src/pages/DividendsWizard/Step-2/TaxWithholdingModal.tsx
```diff
@@ -64,6 +74,12 @@ export const TaxWithholdingModal: FC<Props> = ({
     );
 
     const alreadyExists = matchingIndex !== -1;
+    const isExcluded = includes(exclusionList, valueAddress);
+
+    if (isExcluded) {
```
monitz87: Shouldn't this display a validation error in the address field instead of just closing the form with no indication that the address will not be shown in the table?

RafaelVidaurre: We don't know what the expected behavior is for this. I just prevented the addresses from being added for now, but left Thomas a questions so that he can decide what we should show here. Maybe an alert is better for example instead of invalidating the whole file

RafaelVidaurre:
Closing this PR since this is being remade on a separate branch

PR #529, opened 2019-03-21
chore(new-shared): update formatter to use fixed decimal places

sajclarke:

## This PR: 

Addresses the issue: [Dividends configuration screen 'Approving' rounds up to the first decimal](https://app.asana.com/0/951808452757493/1114659192753397)


RafaelVidaurre:
@sajclarke this is a big change since toTokens is used everywhere. Are we sure this is not breaking anything somewhere else?

pabloruiz55:
Let’s not merge this if in doubt since this is not critical.

sajclarke:
> @sajclarke this is a big change since toTokens is used everywhere. Are we sure this is not breaking anything somewhere else?

This change was requested as per the linked task. When toTokens is used, it's to display the value and doesn't actually manipulate the value so it shouldn't break anything. Depends on if we wish to round up or truncate when displaying values.

packages/new-polymath-shared/src/formatters/index.ts
```diff
@@ -76,10 +76,7 @@ export const toTokens = (
     num = new BigNumber(`${value}`);
   }
 
-  return num
-    .precision(18)
-    .decimalPlaces(decimals)
-    .toFormat();
+  return num.precision(18).toFixed(decimals);
```
Grsmto: That's a tricky one, here you loose the capabilities of the `toFormat()` function which was transforming things like `500000` to `500,000`.

RafaelVidaurre:
I'd make this  change only on the place where it is required.  So basically keep the `toTokens` as it is and just pass a `decimals` value to it

packages/new-polymath-shared/src/formatters/index.ts
sajclarke: Agreed. This was oversight on my part.

sajclarke:
> Changes requested as this PR would break formatting elsewhere

I've made changes so that the toTokens function remains intact but now it rounds down to preset decimal places

RafaelVidaurre:
@sajclarke two things:

- What does that second number do? Is it some default? (Because we have the default in the function arguments and it is set to one. Why not just pass the decimals value in the modal component instead of changing the formatter itself?

sajclarke:
> @sajclarke two things:
> 
> * What does that second number do? Is it some default? (Because we have the default in the function arguments and it is set to one. Why not just pass the decimals value in the modal component instead of changing the formatter itself?

The decimals value is already being passed to the `toTokens` method. The second number sets the Rounding Mode for BigNumber calculations. A rounding mode of 3 will round down as per [bignumber.js docs](https://mikemcl.github.io/bignumber.js/#dp)

> * I see you managed to commit a piece of code that was not formatted by prettier. How are you performing commits? We have a pre-commit hook that enforces code to be correctly formatted

I'm not sure which commit you're referring to but I'm following the guidelines on the ReadMe afaik

RafaelVidaurre:
@sajclarke about the commits:

`return num.precision(18).decimalPlaces(decimals,3).toFormat();`

This line should've been formated to:

`return num.precision(18).decimalPlaces(decimals, 3).toFormat();`

The fact that it didn't means that somehow neither your editor nor the `pre-commit` hook set up in the repository are running in your machine. That's unrelated to this issue so we can discuss this internally though.

About the rounding mode. Please check if Bignumber.js exports a constant for the rounding mode and use that instead of a magic number. If it doesn't export a constant then let me know and we can just merge this, but I really hope it does ;p

sajclarke:
> About the rounding mode. Please check if Bignumber.js exports a constant for the rounding mode and use that instead of a magic number. If it doesn't export a constant then let me know and we can just merge this, but I really hope it does ;p

@RafaelVidaurre  Yes it does. I've added that change in the latest commit

PR #552, opened 2019-03-26
fix(new-issuer): confirm unposted transactions

BilalAlGhazi:

## This PR:

Fixes the bug [Issuer gets no warning when navigating backwards from dividends step 2 to dividends step 1 while having not submitted investors](https://app.asana.com/0/951808452757493/1115263376778667)

## Notes

If there are transactions that are not posted to the blockchain:

- If the user clicks on `GO BACK` from the wizard, the modal confirm will be displayed.
- If the user clicks on the `refresh` button on the browser or types in a new URL, a native confirm modal will be shown, the text of this modal can not be customized as per [this reference](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Notes). Support for custom text has been removed. So the message bellow will be displayed:
Screen Shot 2019-03-26 at 5 53 24 AM


RafaelVidaurre:
Will merge after conflicts are solved @BilalAlGhazi

PR #557, opened 2019-03-27
fix(new-issuer): wizard url direct access

BilalAlGhazi:

## This PR:

Fixes the issue [Issuer can directly access URL to create dividends even if they are disabled](https://app.asana.com/0/951808452757493/1113479795804217)

### DIFF
--- a/packages/new-polymath-issuer/src/pages/DividendsWizard/Container.tsx
+++ b/packages/new-polymath-issuer/src/pages/DividendsWizard/Container.tsx
@@ -6,6 +6,8 @@ import { DataFetcher } from '~/components/enhancers/DataFetcher';
 import {
   createTaxWithholdingListBySymbolAndCheckpointFetcher,
   createCheckpointBySymbolAndIdFetcher,
+  createCheckpointsBySymbolFetcher,
+  createDividendsByCheckpointFetcher,
 } from '~/state/fetchers';
 import { types, formatters, utils } from '@polymathnetwork/new-shared';
 import { DateTime } from 'luxon';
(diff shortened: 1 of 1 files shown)


packages/new-polymath-issuer/src/pages/DividendsWizard/Container.tsx
```diff
@@ -203,48 +208,118 @@ export class ContainerBase extends Component<Props, State> {
...
+                      map(dividendsData, dividends =>
+                        map(dividends, dividend => {
+                          const { expiry, investors } = dividend;
+
+                          const remainingPayments = investors.filter(
+                            (investor: any) =>
+                              !investor.paymentReceived && !investor.excluded
+                          ).length;
+
+                          const remainingTransactions = Math.ceil(
```
monitz87: Calculating this value is unnecessary (so importing `DIVIDEND_PAYMENT_INVESTOR_BATCH_SIZE` is unnecessary as well). As long as there is 1 remaining payment, we should redirect)

packages/new-polymath-issuer/src/pages/DividendsWizard/Container.tsx
```diff
@@ -203,48 +208,118 @@ export class ContainerBase extends Component<Props, State> {
...
-                fetchBalance={this.fetchBalance}
-                fetchIsValidToken={this.fetchIsValidToken}
-                isLoadingData={loading}
+              <DataFetcher
+                watchProps={this.state}
+                fetchers={fetchers}
+                render={(dividendsData: {
+                  [key: string]: types.DividendEntity[];
+                }) => {
+                  const allDividendsCompleted = every(
```
monitz87: For readability, I'd suggest avoiding this level of nesting (considering that this is already a DataFetcher within another DataFetcher). I'd flatten the array first, store it in a variable, then call `every` on that variable and simply return the `completed` calculation in the callback

BilalAlGhazi: I tried to do it without 2 nested `map` statements, but could not, because we have an object that has multiple arrays, each array has an array of investors, so the data is actually tucked in the third level.

monitz87: `flatten` should take care of that by itself without the need for `map`

RafaelVidaurre:
Yup, this can't be done in any other way since handlers only exist in modern

PR #559, opened 2019-03-29
Feature/currency select disabled tooltip

Grsmto:

### DIFF
--- a/packages/new-polymath-issuer/src/pages/DividendsWizard/Step-3/index.tsx
+++ b/packages/new-polymath-issuer/src/pages/DividendsWizard/Step-3/index.tsx
@@ -127,6 +127,12 @@ const Step3Base: FC<Props> = ({
 
   const [erc20TokenSymbol, setErc20TokenSymbol] = useState('');
 
+  const isTestNet = [
+    constants.NetworkIds.Kovan,
+    constants.NetworkIds.Local,
+    constants.NetworkIds.LocalVm,
+  ].includes(networkId);
+
   useEffect(
     () => {
       const { isSubmitting, submitEvent } = formSubmissionStatus;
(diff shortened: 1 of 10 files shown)


packages/new-polymath-issuer/src/pages/DividendsWizard/Step-3/index.tsx
```diff
@@ -343,14 +343,34 @@ const Step3Base: FC<Props> = ({
...
-                        types.Tokens.Usdt,
+                        {
+                          value: types.Tokens.Erc20,
+                        },
+                        {
+                          value: types.Tokens.Dai,
+                        },
+                        {
+                          value: types.Tokens.Gusd,
+                          isDisabled: networkId === constants.NetworkIds.Kovan,
```
monitz87: This should be stored into a variable instead of doing the comparison in every item

packages/new-polymath-issuer/src/pages/DividendsWizard/Step-3/index.tsx
```diff
@@ -343,14 +343,34 @@ const Step3Base: FC<Props> = ({
...
+                        },
+                        {
+                          value: types.Tokens.Pax,
+                          isDisabled: true,
```
monitz87: Why is this always disabled?

Grsmto: Typo! :o

packages/new-polymath-issuer/src/pages/DividendsWizard/Step-3/index.tsx
Grsmto: Used the already existing `isTestNet` instead of this.

PR #560, opened 2019-03-29
Fix/dividends details page update

BilalAlGhazi:

## This PR:

Fixes the issue: [Dividends details screen doesn't get immediately updated after submitting the transactions](https://app.asana.com/0/951808452757493/1115334145499354)


packages/new-polymath-issuer/src/pages/DividendDetails/Presenter.tsx
```diff
@@ -465,9 +476,15 @@ export const Presenter = ({
...
-      </Heading>
+      {loading ? (
+        <Box mt="l" mb="m">
+          <LoadingDots />
+        </Box>
+      ) : (
+        <Heading variant="h2" mt="l" mb="m">
+          Transactions
+        </Heading>
+      )}
```
Grsmto: Why is there a loading for this? Shouldn't it hide the entire table instead of just the heading?

Grsmto: Why is there just the title being replaced by the loader? Shouldn't it be the entire table instead?

BilalAlGhazi: I did the loading on the title to go with the cards at the top, but I can change that to replace the table with the loading indicator instead

Grsmto: Not sure to understand. The title doesn't change/load anything. So I don't get why there should be a loading there at all! 😅

BilalAlGhazi: I changed the loading indicator to replace the transactions table instead of the title.

PR #596, opened 2019-06-20
Refactor/3.0 contracts

remon-nashid:
### packages/new-polymath-scripts
 - Now includes latest fixtures.
	- 2.x ABIs, if needed, are left in `src/fixtures/contracts/2.x`.

###  packages/new-polymath-sdk
- Synced `packages/new-polymath-sdk` with `polymath-sdk#beta` branch. Initially, I was aiming to replace only required parts (namely dividend-related features). However, the interdependency between sdk files made it impossible to pick certain files. I ended up just replacing `new-polymath-scripts/src` with `polymath-sdk` files. 
- Even the change above couldn’t be done as clean as I hoped. External SDK packages has been growing past the internal one. It’s not quite a drop-in replacement. I made the following large scale code replacement to make the later work easily with `new-polymath-issuer`.
    - Renamed all instances of `securityTokenSymbol` to just `symbol`. Also renamed `CheckpointIndex` to `CheckpointId`. The later just represents smart contract `CheckpointID` ie an auto incremented integer.
    - To simplify things, reliance on uuid has been removed in favour of compound keys (eg symbol + checkpointID…etc).

### packages/polymath-issuer
- Webpack dev and production config have been modified slightly. For some reason, a `Cyclic dependency` issue occurred once I copied `poly-sdk` source files to `new-poly-sdk`. The work around was to add `chunksSortMode: 'none’` option to `HtmlWebpackPlugin`, per https://github.com/jantimon/html-webpack-plugin/issues/981
- “Token Name” has been moved from Ticker reservation to token creation screen.

### packages/polymath-js
- This is where the bulk of `3.0` changes have taken place. Similar to the SDK, SecurityToken should be instantiated via a static function called `create()`. The later will detect deployed contract version and load relevant ABI.
- The signature of contract wrapper methods have been changed according to 3.0. In case the ST being manipulated is of an older version, method calls will fallback to the 2.0 specific method. See `transferManager.modifyKYCDataMulti(whitelistItems)` for example.
- When relevant, Interface ABI (eg IST or ISTR) will be used, exclusively, instead of the implementation one.

## Paths you probably don’t need to review:
- `packages/new-polymath-sdk/src/LowLevel/` is similar to the corresponding one in `polymath-sdk`
- `packages/new-polymath-scripts/src/fixtures/` These fixtures have been created on `apps` repo (to be merged via https://github.com/PolymathNetwork/apps/pull/149). Then I copied those fixtures over to this repo

## How to review
- Apparently this is not doable via Github UI. To get an overview of the files changed, pull PR branch locally, and diff it against develop as follows. This command excludes noise: 
```
git diff origin/develop --name-status -- . ':!packages/new-polymath-scripts/src/fixtures/' ':!packages/new-polymath-sdk/src/LowLevel/' ':!yarn.lock'
```
- To see the actually diffs just remove `—name-status` option from the command above.

In brief, let’s review this PR over a call.


packages/new-polymath-issuer/src/state/sagas/transactionQueues/index.ts
```diff
@@ -36,6 +36,7 @@ export function* runTransactionQueue<Args, ReturnType>(
 
   const { uid } = transactionQueue;
 
+  // @ts-ignore
```
monitz87: Why is this necessary?

packages/new-polymath-sdk/src/LowLevel/types.ts
```diff
@@ -29,7 +29,7 @@ export interface GenericContract {
 
 export interface Erc20DividendDepositedEvent {
   _depositor: string;
-  _checkpointId: string;
+  _checkpointId: number;
```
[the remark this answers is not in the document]
remon-nashid: That was unnecessary. Resolved.

packages/new-polymath-issuer/src/state/sagas/transactionQueues/index.ts
remon-nashid: Not anymore.

remon-nashid:
Ready for another review.

PR #628, opened 2019-07-29
fix: better error display + stage providers' maintenance page

remon-nashid:
- Another attempt at catching and printing sendgrid errors.
- Added ProvidersMaintenancePage. To be used later if needed.


remon-nashid:
@FahdW we'll replace `ProviderPage` in `routes` with `ProvidersMaintenancePage`. Nothing fancy. But we won't do that now as it will hinder our ability to test.

PR #637, opened 2019-08-16
Chore: alter provider's app form text on testnets

remon-nashid:


packages/polymath-issuer/src/pages/providers/ApplyForm.js
```diff
@@ -65,7 +65,8 @@ const linkTooltip = (title: string) => (
   </Tooltip>
 );
 
-export const ApplyFormComponent = ({ handleSubmit, onClose }) => {
+export const ApplyFormComponent = args => {
```
FahdW: change args for props, since this is a component

PR #639, opened 2019-08-21
chore(issuer): providers:: add DWF to legal section

remon-nashid:


remon-nashid:
Yup, offchain-production record is in place.

PR #645, opened 2019-09-04
fix(issuer): fix network check

remon-nashid:


remon-nashid:
> Why not use a non type check? ==?

Because it adds a tiny bit of uncertainty.

PR #646, opened 2019-09-23
Feature/individual restrictions

FahdW:


remon-nashid:
I found that when I set token amount to `0`, form validation would accept it but of course, transaction will revert.

FahdW:
Yeah we can set a minimum of 1 on the token field there so it can prevent that on the form field level.

remon-nashid:
2) we need to validate that token amount is > 0. We need proper validation for percentage type too.

remon-nashid:
3) times should be converted to seconds before being passed to the smart contracts, and then converted to milliseconds after loading them from smart contracts. The reason is that all time validations in solidity, are comparing input dates to `block.timestamp`, which is in seconds. See https://solidity.readthedocs.io/en/latest/units-and-global-variables.html

packages/polymath-issuer/src/pages/restrictions/components/IndividualRestrictionsTable.js
```diff
@@ -0,0 +1,219 @@
...
+
+  formatCell = (id, cell) => {
+    const { individualRestrictions } = this.props;
+    if (cell.value === undefined) return '-';
+    switch (cell.info.header) {
+      case 'customStartTime':
+      case 'dailyStartTime':
+      case 'customEndTime':
+      case 'dailyEndTime':
+        return moment.unix(cell.value / 1000).format('MMM DD YYYY');
```
remon-nashid: If `cell.value` is brought from VRTM module, it would be in seconds. Perhaps we need to _multiply_ it by `1000` here.

FahdW: The reason it needs to be divided by 1000 cause in redux it is stored with it's * 1000 value. It helps fill the time stamp, otherwise it wouldn't work.

remon-nashid:
Also, I'm not sure how can I pick time later than 10:00, because remaining times are chopped:

Screen Shot 2019-09-30 at 5 38 42 PM

FahdW:
I had to add that restriction to stop allowing time in the past because now it's preventing submission if the date is in the past. So the form needs to reflect this restriction.

remon-nashid:
Hey @FahdW, I found that this issue still persists:
> we need to validate that token amount is > 0. We need proper validation for percentage type too.
The form would let me submit, but the ethereum transaction would revert. **Note** that I reproduced this with individual transactions, but it might apply to other types, too.

remon-nashid:
**This one applies to all kinds of restrictions**

Attempting to _configure_ a restriction that has already started (start datetime in the past), transaction would fail. Perhaps we need to prevent modifying a restriction past its start date.

remon-nashid:
Clicking anywhere in an  _empty_ individual restrictions page, would crash the dapp, due to an uncaught exception ```IndividualRestrictionsForm.js:474 Uncaught TypeError: Cannot read property 'dailyStartTime' of undefined```

ndvd-table

remon-nashid:
Everything other than the issues above looks good.

remon-nashid:
So we have two issues left from the above:
- Inabitlity to choose hours later in the day (9:00 pm and up) because the select list is chopped.
- Restrictions shouldn't be editable once they've became effective.

FahdW:
You can select hours beyond 10, you have to scroll the modal. Not ideal, but sort of what happens in a scrolling modal.

Restrictions are editable, but they will be flagged to change the start time since modifying it means essentially recreating it.

remon-nashid:
- That's what I mean, I cannot scroll further than 10pm, overflow will eat remainin options.
- Editing a running restriction actually results in a revert and transaction fails.

FahdW:
Yeah, there's validation that stops from proceeding if the start date is in the past.

PR #650, opened 2019-10-08
add vertalo, finoa, froriep and cap. table management

FahdW:


FahdW:
We may need to rework some of this, when testing "I have my own" causes the tabs to spill over. It's also why there is so much space between each tab. We may need to rework that provider page.

remon-nashid:
Oh no need.

PR #652, opened 2019-10-09
Fix for forever running transactions after metamask rejection

remon-nashid:
Just added a catch-all exception handler. It will stop the spinner in case of exceptions in general, instead of just detecting and handling a few.


remon-nashid:
Thanks @FahdW

PR #660, opened 2019-10-31
GT-94: notify users about csv changes

remon-nashid:


remon-nashid:
@FahdW build checks are successful now.

PR #663, opened 2019-11-07
WIP: Feat/manual approval

FahdW:


packages/polymath-issuer/src/pages/compliance/components/ApprovalTable.js
```diff
@@ -0,0 +1,238 @@
...
+    switch (cell.info.header) {
+      case 'fromAddress':
+      case 'toAddress':
+        return addressShortifier(cell.value);
+      case 'expiry':
+        return moment.unix(cell.value).format('MMM DD YYYY, hh:mm:ss a');
+      case 'tokens':
+        return (
+          <a
+            href={`https://kovan.etherscan.io/tx/${row.cells[3].value}`}
```
FahdW: Need to modify this

remon-nashid:
@FahdW I updated MATM abi one more time to catch up with a few more changes.

packages/polymath-issuer/src/pages/compliance/components/ApprovalTable.js
```diff
@@ -0,0 +1,238 @@
...
+  {
+    header: 'To Investor Wallet Address',
+    key: 'toAddress',
+  },
+  {
+    header: 'Approval Expiry',
+    key: 'expiry',
+  },
+  {
+    header: 'Unique ID (TxHash) of Approval',
```
remon-nashid: I don't see this `Unique ID` column displayed. I'm not even sure if it's necessary TBH

Screen Shot 2019-11-08 at 4 15 54 AM

packages/polymath-js/src/contracts/ManualApprovalTransferManager.js
```diff
@@ -0,0 +1,85 @@
...
+    const approvals = await this._methods.getAllApprovals().call();
+    const formattedApprovals = [];
+
+    for (let i = 0; i < approvals['0'].length; i++) {
+      let a = {};
+
+      a.id = approvals['0'][i] + approvals['1'][i];
+      a.fromAddress = approvals['0'][i];
+      a.toAddress = approvals['1'][i];
+      let events = await this._contractWS.getPastEvents('AddManualApproval', {
```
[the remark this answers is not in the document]
remon-nashid: If this is for `txHash` then I'd just ignore it. We filter out from the table display anyway.

packages/polymath-issuer/src/pages/compliance/components/EditApprovalForm.js
```diff
@@ -0,0 +1,250 @@
+// @flow
+
```
remon-nashid: Out of curiosity, could we've used one form component for both adding and editing, white disabling addresses fields in the later case?

packages/polymath-issuer/src/pages/compliance/components/EditApprovalForm.js
```diff
@@ -0,0 +1,250 @@
...
+    .string()
+    .isRequired(REQUIRED_MESSAGE)
+    .isAddress(ADDRESS_MESSAGE),
+  toAddress: validator
+    .string()
+    .isRequired(REQUIRED_MESSAGE)
+    .isAddress(ADDRESS_MESSAGE),
+  token: validator
+    .number()
+    .moreThan(0, MORE_THAN_MESSAGE)
```
remon-nashid: I believe `allowance` can be set to zero while modifying an existing approval.

packages/polymath-issuer/src/pages/compliance/components/ApprovalTable.js
FahdW: It didn't let me, kept crashing when I used thst function.

packages/polymath-issuer/src/pages/compliance/components/ApprovalTable.js
FahdW: It's hidden so the data is available when you click Approved token transfer. Goal is to not have to search an array every single time

packages/polymath-issuer/src/pages/compliance/components/EditApprovalForm.js
FahdW: I did this last time for VRTM and it became such a mess it was better to have them a separated concerns.

packages/polymath-issuer/src/pages/compliance/components/EditApprovalForm.js
FahdW: That is fine in the case that someone doesn't want to change their token value. Since this is incremental and not a hard value

packages/polymath-js/src/contracts/ManualApprovalTransferManager.js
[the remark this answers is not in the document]
FahdW: It's for txHash and we use it on the table. So transactions are necessary

packages/polymath-issuer/src/pages/compliance/components/EditApprovalForm.js
FahdW: Just realized I changed how this works. We may need to restrict that part

remon-nashid: I don't follow. If one doesn't want to change the value, they'll leave it as `0`. But the validation above prevents that.

remon-nashid:
Some tests are failing.

packages/polymath-issuer/src/pages/compliance/components/AddApprovalForm.js
```diff
@@ -202,21 +202,22 @@ const formikEnhancer = withFormik({
...
   handleSubmit: (values, { errors, setFieldError, props }) => {
     const { dispatch, handleClose, approvals } = props;
-    const approvalExists = approvals.find(
-      i =>
-        i.fromAddress === values.fromAddress && i.toAddress === values.toAddress
-    );
-    if (approvalExists) {
-      setFieldError(
-        'toAddress',
-        'An approval with both addresses already exists'
```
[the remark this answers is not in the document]
FahdW: Needs to be reworked, the approvals that expire keep being fetched. Unfortunately because of this the validations would prevent a user from being able to apply a new approval.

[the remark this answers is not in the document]
FahdW: Updated the validation logic @remon-nashid

remon-nashid:
@FahdW could you please review all the issues above, resolve them, then ping me again for a re-review?

packages/polymath-issuer/src/pages/compliance/components/EditApprovalForm.js
FahdW: They wouldn't leave it as 0. They would leave it as the value of the original. for eg, if a user has entered 100 Tokens, the edit form will fill it with 100 Tokens. It will mark the difference and select the correct adjustment for the token value

PR #682, opened 2019-12-17
GT-152 increase gasLimit for certain tx

remon-nashid:


remon-nashid:
Failing test is successful now. Ready for review.

PR #689, opened 2020-01-10
GT-163: stop calling etherscan api

remon-nashid:
Remove the call to etherscan on every visit to the dashboard.


remon-nashid:
> It's yarn that's failing, just pull develop down

This is a hotfix to master.

FahdW:
My mistake, yeah realized shortly after. Just add 1.21.1 to yarn, or 1.x at this point.

remon-nashid:
Yup, should be good now.

PR #707, opened 2020-02-10
add equa and move bitcoin suisse

FahdW:


packages/polymath-issuer/src/pages/providers/data.js
```diff
@@ -247,7 +247,7 @@ const providers: Array<ServiceProvider> = [
   },
   {
     id: 56,
-    cat: 0,
+    cat: 4,
```
remon-nashid: shall we move related images to `custody` directory?

remon-nashid: I realize it should work as-is, but just to prevent confusion/issues in the future.

PR #709, opened 2020-02-19
Feat/manual approval

FahdW:

### DIFF
--- a/packages/polymath-issuer/src/actions/compliance.js
+++ b/packages/polymath-issuer/src/actions/compliance.js
@@ -1168,13 +1168,7 @@ export const editManualApproval = (
   const editingApproval = getState().whitelist.editingApproval;
   const newAllowance = new BigNumber(Web3.utils.fromWei(allowance).toString());
   const oldAllowance = new BigNumber(editingApproval.tokens.toString());
-  let increase;
-
-  if (newAllowance.isGreaterThan(oldAllowance)) {
-    increase = true;
-  } else {
-    increase = false;
-  }
+  const increase = newAllowance.isGreaterThan(oldAllowance);
   const changeAmount = Math.abs(oldAllowance.minus(newAllowance));
 
   dispatch(
--- a/packages/polymath-issuer/src/actions/compliance.js
+++ b/packages/polymath-issuer/src/actions/compliance.js
@@ -80,9 +80,9 @@ export const editApproval = approval => ({
   approval,
 });
 
-export const MODIFY_APPROVALS = 'compliance/MODIFY_APPROVALS';
-export const modifyApprovals = approval => ({
-  type: MODIFY_APPROVALS,
+export const MODIFY_APPROVAL = 'compliance/MODIFY_APPROVAL';
+export const modifyApproval = approval => ({
+  type: MODIFY_APPROVAL,
   approval,
 });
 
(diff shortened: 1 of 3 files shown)


packages/polymath-issuer/src/actions/compliance.js
```diff
@@ -1166,10 +1174,10 @@ export const editManualApproval = (
...
   const st: SecurityToken = getState().token.token.contract;
   const editingApproval = getState().whitelist.editingApproval;
-  const increase = editingApproval.tokens < allowance ? false : true;
-  const changeAmount = Math.abs(
-    editingApproval.tokens - Web3.utils.fromWei(allowance)
-  );
+  const newAllowance = new BigNumber(Web3.utils.fromWei(allowance).toString());
+  const oldAllowance = new BigNumber(editingApproval.tokens.toString());
+  const increase = newAllowance.isGreaterThan(oldAllowance);
+  const changeAmount = Math.abs(oldAllowance.minus(newAllowance));
```
remon-nashid: Use `BigNumber` `absoluteValue()` instead of Math.

remon-nashid: What happens if change amount is zero? Will it still send a transaction?

FahdW: If it's 0 it will still send, as a user maybe changing the date value which is perfectly valid.

packages/polymath-issuer/src/actions/compliance.js
```diff
@@ -1147,7 +1147,15 @@ export const addManualApproval = (
...
         };
-        dispatch(addApproval(approval));
+
+        let approvalIndex = getState().whitelist.approvals.findIndex(
+          i => i.id === approval.id
+        );
+        if (approvalIndex === -1) {
+          dispatch(addApproval(approval));
+        } else {
+          dispatch(modifyApprovals(approval));
```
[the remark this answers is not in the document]
FahdW: I put approvals cause it modifies the entire list and not just append. But it's not a big deal

packages/polymath-issuer/src/actions/compliance.js
remon-nashid: right!

packages/polymath-issuer/src/actions/compliance.js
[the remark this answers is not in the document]
remon-nashid: ahh ok.

packages/polymath-js/src/contracts/ManualApprovalTransferManager.js
```diff
@@ -63,20 +63,40 @@ export default class ManualApprovalTransferManager extends Contract {
...
+        {
+          filter: { _from: a.fromAddress, _to: a.toAddress },
+          fromBlock: 0,
+          toBlock: 'latest',
+        }
+      );
+
+      if (
+        modifyEvents.length > 0 &&
+        modifyEvents[modifyEvents.length - 1].blockNumber >
```
remon-nashid: What's the purpose of this condition? Is it to get txHash of the modification event (if any), otherwise get txHash of  the approval event? Are we sure that events are sorted by block number, and ascendingly? Otherwise, the last event of in the array might not be the event we're looking for.

Why are we storing txHash at the first place?

I would merge both events arrays, sort them by block number, and then select last element of that merged array.

FahdW: The last value in the array is always the latest event from the getPastEvents result. The hash needs to be stored because it is clickable on the table to get to etherscan. Merging both arrays is not necessary since you will have excess data that you do not need to go through. You only need the last of the two events.

PR #724, opened 2020-04-06
Chore/gt 183 sp copy changes

remon-nashid:
just some boring copy changes to SP page.


packages/polymath-issuer/.env.local
```diff
@@ -42,4 +42,5 @@ REACT_APP_POLYMATH_REGISTRY_ADDRESS_LOCAL=0x9FBDa871d559710256a2502A2517b794B482
 REACT_APP_DEPLOYMENT_STAGE=local
 REACT_APP_POLYMATH_REGISTRY_ADDRESS_KOVAN=0x5b215a7d39ee305ad28da29bf2f0425c6c2a00b3
 REACT_APP_POLYMATH_REGISTRY_ADDRESS_GOERLI=0x7e3c8aF98538Ba19A10Dfc7E8F5469a76998b0f0
-REACT_APP_ALLOW_GANACHE_ONLY=false
\ No newline at end of file
+REACT_APP_POLYMATH_REGISTRY_ADDRESS_MAINNET=0xdfabf3e4793cd30affb47ab6fa4cf4eef26bbc27
```
FahdW, line 45: Any reason we need to include MAINNET address?

remon-nashid: Any reason we shall not? It was never tracked and I thought I'd add it here. Useful for testing.

FahdW: This will enable mainnet on betastudio.

remon-nashid: Not without configuring offchain as well. Anyway I'll remove it.

PR #743, opened 2020-06-23
Heap to PROD

honarpour:


FahdW:
For this we need usually go dev -> release -> prod

PR #771, opened 2020-08-20
add functionality for preminting

FahdW:


remon-nashid:
@FahdW While the UI surface of this PR is small, it brings non-trivial and perhaps breaking changes to STOs. Can we spin up a heroku instance to get a couple of hands testing this change, before we merge it into develop? Concerns are around  and

